PR #14, opened 2015-05-06
searching for phone numbers

geier:
with new command 'telephone' (I know that name is far from optimal, feel free to change it)

### DIFF
--- a/khard/config.py
+++ b/khard/config.py
@@ -36,10 +36,10 @@ class Config:
             if self.config['general'].has_key("default_action") == False:
                 print "Error in config file\nMissing default action parameter."
                 sys.exit(2)
-            elif self.config['general']['default_action'] not in ["list", "details", "new", "modify", "remove", "mutt", "twinkle", "alot", "source"]:
+            elif self.config['general']['default_action'] not in ["list", "details", "new", "modify", "remove", "mutt", "twinkle", "alot", "source", "telephone"]:
                 print "Error in config file\n" \
                         "Non existing value for default action parameter\n" \
-                        "Possible values are: list, details, mutt, twinkle, alot, new, modify, remove and source"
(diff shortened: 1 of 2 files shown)


scheibler:
What exactly do you want to achieve? At the moment you print all phone numbers from your address book
to a single output list.

Here are some things, which you already can do with khard:

If you want to list all phone numbers of a contact, type:
khard -s anton details

If you want to search for a specific phone number:
khard -s 01234567890 details
or
khard -s "0123 4567890" details
or
khard -s 56789 details

And if you only want to know who belongs to a given phone number:
khard -s "0123 4567890" twinkle
Output: Anton (cell)

Maybe the option name "twinkle" is a bit confusing. I've chosen it, cause that's my used SIP client.
Could be better to rename it to "phone" or something more clearly.

Could you provide further information about your motivation please?

Cheers
Eric

geier:
I want what `mutt` and `alot` are for email addresses, but for telephone numbers (as I often initiate phone calls from the command line, via [frytz](https://github.com/geier/frytz/).

If I understand twinkle right, it is a kind of *reverse* search for telephone numbers, right?

scheibler:
I understand. Yes, the twinkle option was some kind of reverse phone number search. With the latest
commit I renamed twinkle to phone. Now the phone option does the same for phone numbers like mutt
for e-mails (as you requested). The reverse phone number search moved directly into the twinkle
plugin. So it's a more general approach. Now every phone client can process khard's output on it's
own.

The output lines have the following structure:
number<TAB>name<TAB>label

It's the same like mutt's preferred output format: One phone number per line, separated by
tabulators. Therefore it should be easy to parse. It quits with exit code 1 if no phone number was
found. Please tell me, if this fits your needs.

BTW Your fritz box caller script sounds like a cool idea. I will try it.

Cheers

geier:
Thanks!

PR #22, opened 2015-08-20
New option "merge".

kmq:
Hi, 

I added an option to merge contacts.
Basically is opens them like in 'modify' in a merge_editor (like vimdiff) and then updates the first one and removes the second.

Thank you for this very useful tool,

Regards,
kmq


coachshea:
This sounds like a great idea. I can't wait to see it in action.

PR #28, opened 2015-10-09
Add Zsh completion definition

okapia:
This adds a file for zsh completions so zsh users will get options, actions and addressbooks completed after khard instead of filenames. Would be nice if it could be included.


scheibler:
Awesome. Thank you for sharing. Works very well. Especially the address book auto completion is great.

PR #52, opened 2016-02-19
Move subcommand handling to argparse.

lucc:
The idea behind this is that we can get detailed help messages from argparse and that we can tell argparse to accept different options for different subcommands.  So in the end I hope to reduce the command line parsing and validating logic in main() and get a nicer command line interface.

Are you interested in merging this?  (I open this PR early in order to get feedback and start a discussion, the work is not done yet.)

TODO:

- [x] move parsing of actions to subparsers
- [x] add help for subparsers
- [x] only accept valid options with different subparsers (partly done)
- [ ] use a function for each subcommand to clearly separate the logic of the different subcommands (this will most likely change the command line interface a little)
  - [x] `email`
  - [x] `phone`
  - [x] `new`
  - [x] `add-email`
  - [x] `modify`
  - [x] `copy`
  - [x] `move`
  - [x] `list`
  - [x] `source`
  - [ ] `export`
  - [ ] `details`
  - [x] `merge`
  - [x] `remove`
- [ ] maybe more ...

BUGS:
- [x] it is not possible to run khard without a subcommand


scheibler:
Yes, I'am very interested in that. Especially after the new args -t and -u a cleanup of the command
line interface would not hurt. I hat this on my to-do list but never found the time to work it out.

For example we could add a special help text for the -s / --search option, depending on the selected
action. At the moment I've specified the different syntax for copy / move, merge and the rest in one
help text (line by line). Furthermore we should consider to separate the -t option into -i /
--input-file for new and modify and -o / --output-file for the export action.

lucc:
Maybe we can also discuss some changes that might be necessary or easy to make with this. At least I would like to hear your opinion on these changes while this is still WIP. For example 2469664.

My aim is to have a function for each subcommand which will be selected based on the `args` object. In the best possible solution these functions should be written in a way that they depend only on their arguments and don't interrogate config options or such themselves (kind of like in functional programming). But I think I have to make some drastic changes to the main function. Do you think that is OK?

lucc:
This has now reached a state where you can test it. I changed the command line interface in c5c3004 (introduce subcommands), 5aed2cf, 47ee280, e3fd6cf, 8a9d9ff, fdd9440 (minor option changes) and especially 7e1e0f3 (main interface change). So you might want to have an extra look at these commits.

I'm happy about comments, otherwise I think we should test it a little, then it can be merged.

lucc:
I'm thinking about fixing the bug that was introduced with this: Currently the default command is not honored. The program just fails (gives a help message) if no command was specified.

It seems that this can easily be fixed in python 3.something. Do you plan to move to python3 soon? Else I will implement a fix similar to [this](http://stackoverflow.com/a/26379693).

scheibler:
Thank you for your work. That looks really promising. Nevertheless, some suggestions:

1. We should rename the template_file_parser to input_file_parser and the -t / --template-file
option to -i / --input-file based on the -o / --output-file option for the export command.

2. Default action: The migration to python3 is on my to-do list but I think, that this still will
take some time. So we should try the method from your stackoverflow link.

3. Now we have to abide a strict option ordering. Previously you could do the following:

```
khard -u uid modify
```

or

```
khard modify -u uid
```

Now we have to put the action before the options of a subcommand. But we still can do:

```
khard -a family modify
```

cause it's an option of the main parser. That's a bit inconsistent but I guess, that we have no
influence on that.

4. I'am not sure, if we should make the search a positional argument. Consider the following:

```
khard modify john smith
```

and

```
khard copy john smith
```

For the second command, you have to escape the search terms, cause we await two positional arguments,
naming "source contact" and "target address book". Same for merge.

What about:

```
khard [email|modify|phone|remove|...] [-s|--search-contact "john smith"] [-u|--contact-uid uid]
khard [copy|move] [--search-source-contact "john smith"] [--source-contact-uid uid] [--target-addressbook family]
khard merge [--search-source-contact "john smith"] [--source-contact-uid uid] [--search-target-contact "max smith"] [--target-contact-uid uid]
```

Advantage: it's completely clear, how to enter source and target. Disadvantage: search args differ
between the latter two commands and the first one and some of them are a bit long.

lucc:
About your comments:
1. see new commit 1ff3726
2. will do
3. see dfd2de4, I hope I put everything in the right place and did not forget the option of any subcommand. I also added `-s` as an alias for `--sort` as `--search` is gone (see also 4.)
4. I am against options for the search terms. Most standard unix commands are able to handle the most common argument without an option. So I would like to do it the same way. Also it is less typing.

I will try something like the syntax for `mv` or `cp` with `khard move` and `khard copy`: we can collect all search terms into a list and use the last one for the target address book and all the others to find the source contact. The only thing where spaces would need to be escaped would then be the addressbook name but I think that can not even have spaces. Will try to push a commit later.

lucc:
Point 4 is addressed in 6bb83c9. I noticed that argparse is intelligent enough to put the last argument into the target_search_terms attribute so we do not have to check and split the list ourselves.

The only thing that is left inconsistent at this point is the syntax of the merge command:
```
usage: khard merge [-h] [-a ADDRESSBOOK] [-g] [-r]
                      [-s {first_name,last_name}] [-u UID]
                      [source_search_terms [source_search_terms ...]]
                      target_search_terms
```
We can supply many source but only one target search term as argument (target search terms have to be quoted on the command line to be put together). I will do some more research, maybe I can fix it.

lucc:
Point 2 is addressed in 071f1d6. I am not entirely happy with this as you can now supply command line arguments which do not match a command and these will then be passed to the default subcommand: Set `default_action = list` in the config and the run `khard foobar`. This will add "list" to the command line and in the end run as if you supplied `khard list foobar`. This might be nice for `default_action = list` but I'm not sure if it is a good idea in general.

lucc:
In python 3 this problem with the subcommand might be easier to handle because argparse will return `None` as the selected subcommand if no subcommand was given and will not terminate. So we could leave it like this until we move to python3.

scheibler:
I understand your point and taking the search as positional args would be very convenient but then
you are less flexible. The problem is, that the selection of a contact should be possible per
search, uid or selection from a list equally. That's the difference to the posix cp and mv commands,
there you have only one way to set the source and target files and the optional arguments only
change things like recursive copying (-R).

Nevertheless, for the actions like details, modify, source and so on, that's fulfilled. I can select
the contact for example with:

```
khard modify john smith      # and eventually select from list
khard modify -u uid
khard modify -a familie      # and then select from list
```

But what, if I only want to specify the target address book name and select the source contact from
list?

```
khard copy -a family friends
```

Or if I want to select the source contact by uid and the target contact by search term?

```
khard merge -uid source_contact_uid target contact
```

Maybe we could solve that by escaping every search term, regardless of the specified action. Then
this could be possible:

```
khard merge -uid source_contact_uid "" "target contact"
```

This would fix the inconsistency of the merge action too. But I think, the following is still better:

```
khard [email|modify|phone|remove|...] [-a sbookname] [-u source_uid] [john smith]
khard [copy|move] [-a sbookname] [-u source_uid] [john smith] [--target-addressbook bookname]
khard merge [-a sbookname] [-u source_uid] [john smith] \
    [--target-addressbook tbookname] [--target-contact "max smith"] [--target-uid target_uid] \
```

That's mostly, what you tried to achieve with the smaller compromise at the copy/move and the bigger
one at the merge command. But I think, at least the latter one is not so frequently in use. If you
are a zsh user, you also could use its completion function from misc/zsh/ to simplify typing.

lucc:
I will try to implement your last suggestion for the command line semantics. The places where you have `john smith` in your example should be a single argument or many arguments? (You did quote `"max smith"` so I think you want to have a single argument here.)

scheibler:
Yes, optional arguments always require quoting (as far as I know). So the value for the
--target-contact must be a single arg. Then you can use many arguments for the source contact search
without problems, cause it's the only positional argument left. So john smith doesn't require
quotes.

lucc:
Optional arguments can also collect arbitrary many arguments. See this example:
```python
>>> import argparse
>>> p = argparse.ArgumentParser()
>>> p.add_argument('-o', '--option', nargs='*')
_StoreAction(option_strings=['-o', '--option'], dest='option', nargs='*', const=None, default=None, type=None, choices=None, help=None, metavar=None)
>>> p.add_argument('param', nargs='*')
_StoreAction(option_strings=[], dest='param', nargs='*', const=None, default=None, type=None, choices=None, help=None, metavar=None)
>>> p.parse_args('foo bar baz -o one two three'.split())
Namespace(option=['one', 'two', 'three'], param=['foo', 'bar', 'baz'])
# BUT:
>>> p.parse_args('-o one two three foo bar baz'.split())
Namespace(option=['one', 'two', 'three', 'foo', 'bar', 'baz'], param=[])
```
So I am not sure if this is what we want or not. I will implemnent it with single arguments for the options for now.

lucc:
The commit 548a1d9 should implement the scenario you described at the end of [your post](https://github.com/scheibler/khard/pull/52#issuecomment-189900374).

scheibler:
I guess, that we have to add exceptions for the -d and -v options in the set_default_subparser
function. Besides that I think, that the current state is much more intuitive and straightened up
then before. If you are satisfied too, I would merge the branch in the next days, when I find some
time.

lucc:
Like this?

scheibler:
-v works. -d also is only callable without an action but I think, as long as we don't need the debug
option for a specific action, that's enough too.

lucc:
What do you mean by -v works? I just noticed that we could also add -v [here](https://github.com/lucc/khard/blob/1bf486f240447144291bd9ef781141b1ff58c672/khard/khard.py#L964)

scheibler:
I mean, that the command ¸khard -v` outputs the correct program version now. Without deeper
knowledge of the set_default_subparser function, I also would have added the -d and -v after lines
964 but as far as I can see, your changes are working too. So it's your decision, if you want to
change something again or if we leave it at that.

lucc:
TODO: zsh completion

lucc:
I think this is all for this PR :)

scheibler:
It took a bit longer then anticipated but now I've merged your code into the master branch. I only
fixed some bugs and optimized several help messages (move, copy and merge action).  I'am very happy
with the resulting command line interface. Thank you for your extensive work.

lucc:
Thanks. 

Some questions for further PRs: 

1. Do you care about PEP8? You seemed to have reverted some of my PEP8 fixtures.
2. Do you care about a understandable commit history or do you just commit some stuff every once in a while? I feel that these commits I put here are easier to understand than the giant commit you sqashed it into. I also do not understand all the changes you made ontop of mine as they are hidden in the giant commit.
3. Do you want to split up some big files into smaller files? khard.py could be split into cli.py and subcommands.py to hold the argparse and user interaction stuff (cli) and the plain subcommand functions without further user interactions (subcommands). 
4. Do you care about object oriented programming? I feel that the config class has a lot of responsibilities. Some might be moved to the address book class or so.

I also came up with another idea for the cli interface. It might be interesting to remove the source subcommand and instead add a --source option to edit and merge. `edit --source` would be a direct replacement of `source` and `merge --source` would (obviously) open the two source files in the merge editor.

scheibler:
>1. Do you care about PEP8? You seemed to have reverted some of my PEP8 fixtures.

In general, I care about code structuring and formatting but I must say, that I find the strict
80 chars per line limit a bit ugly sometimes, for example, if you only have one or two chars left.
Nevertheless I was endeavored to follow the PEP8 rules but obviously I missed some line breaks. Or
do you refer to something, I'am not aware of at the moment?

>2. Do you care about a understandable commit history or do you just commit some stuff every once in a while? I feel that these commits I put here are easier to understand than the giant commit you sqashed it into. I also do not understand all the changes you made ontop of mine as they are hidden in the giant commit.

Another thing, that I thought about before merging. On the one hand your commit history for the
feature branch was very detailed and clear and made your code changes very understandable but on the
other hand I didn't want to merge all that small commits into the history of the master branch to
keep it clean too. So I thought about only keep some key commits and squash the other but then I
couldn't decide, which were the important ones. But is the very detailed commit history of one
branch still relevant and useful after some more features were committed to the master branch? Which
behavior do you prefer and what would you suggest for the future?

>3. Do you want to split up some big files into smaller files? khard.py could be split into cli.py and subcommands.py to hold the argparse and user interaction stuff (cli) and the plain subcommand functions without further user interactions (subcommands). 

Good idea.

>4. Do you care about object oriented programming? I feel that the config class has a lot of responsibilities. Some might be moved to the address book class or so.

At the moment, I'am more or less happy with the config class solution. But anyway, which parts do
you have in mind?

>I also came up with another idea for the cli interface. It might be interesting to remove the source subcommand and instead add a --source option to edit and merge. `edit --source` would be a direct replacement of `source` and `merge --source` would (obviously) open the two source files in the merge editor.

Good idea too. Now with the subparsers in place, such changes are much easier to achieve then
before. Should be no big deal to implement that.

lucc:
About commit history: I am used to the way neovim/neovim do things and also have read the [git flow stuff](http://nvie.com/posts/a-successful-git-branching-model) about git history. So when I code for my self or on my projects I tend to keep commits that "tell a story about what is being done" and can still be understand one year later. But this is your project obviously :)

The OO stuff I will explain in further issues and PRs when I come around to do some work there.

PR #58, opened 2016-03-15
python3 compatibility WIP

dmwilcox:
Hello!  And thank you for writing this awesome little project -- it is exactly what I need.  I found myself stumbling over what I realized was my python3 virtual environment this morning and hacked around it.

I can vouch for very basic functionality working in Python3 and it appears work with Python2 even with my changes.  It is hard to know what I might be breaking so it would be great to get some guidance from you.  Specifically regarding testing, edge cases to look out for and if you have a recipe for a regression setup -- to test that older version of vobject you mentioned in the comments.

Let me know what you think, cheers!

### DIFF
--- a/khard/address_book.py
+++ b/khard/address_book.py
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
 
 import os
 
@@ -30,4 +34,3 @@ class AddressBook:
 
     def add_contact(self, contact):
         self.contact_list.append(contact)
-
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -7,7 +7,7 @@ import os, sys, string, datetime, re
 import vobject, yaml
 from atomicwrites import atomic_write
 from pkg_resources import parse_version, get_distribution
-import helpers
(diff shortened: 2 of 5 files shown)


scheibler:
Yes, python3 support definitely is on my todo-list. Together with that, I also want to throw out the
legacy code for the vobject version < 0.8.0 cause the new library version is installable via pypi
now and that would make things much easier. But currently I'am rebuilding the whole console
interface and this brings massive changes especially for the khard.py main class. So if you want to
help with python3 support I advice you to wait, until that's done. Thanks.

dmwilcox:
OK!  Sounds great.  Keep me in the loop.

BTW: the one part that seems to be left out of all the "migrate off your stuff off the cloud" documentation.  Is actually dealing with the giant vcf and ics files a provider like google will give you.  As part of setting up khard I wrote a few little tools that might be helpful to that end.  I'll try and document them and publish them to pip soon-ish.  Let me know what you think, cheers,

=D
https://github.com/dmwilcox/vcard-tools

lucc:
@dmwilcox the interface change from #52 has reached master in d8de85a191c19c60fb689e212ae7487c985c1014. I also started a python 3 port while I was waiting for it to be merged. But I did not care for or test python2 compatibility so it might be quiet different. I will put it up, so we can compare these two approaches.

PR #67, opened 2016-05-09
Correct query_command in mutt documentation

maethor:
This PR correct a typo the mutt section of the documentation. `khard email -p %s` shows a "pretty" display which doesn't work in mutt. `khard email %s` is working just fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ Khard may be used as an external address book for the email client mutt. To acco
 following to your mutt config file (mostly ~/.mutt/muttrc):
 
 ```
-set query_command= "khard email -p %s"
+set query_command= "khard email %s"
 bind editor <Tab> complete-query
 bind editor ^T    complete
 ```


scheibler:
I swapped the meaning of -p from pretty to parsable in the latest khard version 0.10.0 and the
readme file at the github website reflects that change. But I havn't tagged 0.10.0 as a release yet
cause i'am still waiting for feedback. So if you still installed khard version 0.9.0 from the pypi
repository, you have to refer to the readme file, that is shipped with that release and don't use -p
for the mutt command.

I should have used the additional option (-p) for the parsable output in the first place cause it's
much easier to put that once in the mutt config instead of writing that option every time when I
want to view the phone numbers or email addresses of a contact in a pretty formatted table. And
redefining the same parameter to a different function seems not so ideal too. I'am sorry for the
inconvenience.

maethor:
OK, thanks :)

I agree with you on the -p option.

PR #79, opened 2016-07-19
Add --config option to use different config file.

lucc:


scheibler:
Nice idea. But for me it doesn't work. Problem is, that the set_default_subparser function must be
called before parse_args. The set_default_subparser function requires the default action which must
come from the config file. But loading of that config file is not possible cause we don't get the
path of the new config file before parse_args (lines 1459 to 1460 of khard.py).

The only thing, I could imagine is to take the default action from the default config file, parse
args, get path to new config file and recreate config object. Or do I miss something here?

lucc:
That is right. My solution would be #81, alas to remove the `set_default_subparser`. This can only be done now because only python3 can detect a missing subcommand like it is done in #81.

You should merge #81 before this one.

scheibler:
Sorry, that doesn't work.

First, as you've already pointed out, command line options wouldn't work any longer for the default
action. Unfortunately this also includes the indispensable search option. Example: My default action
is phone, cause I want to look up phone numbers with minimal typing like `khard first_name
last_name`. But that would not be possible any longer. The only thing I still could do with the
default action is show all phone numbers in my address book. But then the default action would be
almost useless (in my opinion).

Second: Try to set phone or email as your default action and run khard without any action or command
line option. Then you will get an AttributeError at line 1608. It seems, that additionally added
command line options like --parsable are not respected in that case.

Please also see the commit a67f88a4510d0385e5e7357aa5c7815388eee198 for further information.

lucc:
I implemented the two step parsing. It

* adds a `--config` option
* removes the old argparse patch
* processes `--help` before the config file is read (config errors wont stop the help text from being displayed)
* reads the default command from the correct config file (even wth `--config`)
* arguments to the default command can be given on the command line without giving the default command on the command line
* **feature/bug?** global options can be given after the subcommand, subcommand options can not be given before the subcommand (the parser will think no subcommand was specified and use the default one, the option will probably be used as a search term then, depending on the subcommand)

If you merge this try to keep these commits as one of the commit messages mentiones a commit hash of another (or you have to change the commit message).

scheibler:
Looks really good. Even the default action is processed well. Thanks for reorganizing.

I only found one bug so far. If you run

```
khard -h
```

the general help text is shown but if you run

```
khard phone -h
```

it's also the general help and not the help text for the phone action.

lucc:
I think I fixed it but there is a small (in my opinion acceptable) trade off: Some commits back we could specify global options after the subcommand. Now this can't work anymore b/c it was exactly that which "eat" the `-h` of the subcommand.

scheibler:
I can live with that. Works.

lucc:
I already had tests for this in #96  and didn't notice :(

scheibler:
Merged into develop branch.

lucc:
How do you merge these commits all the time to get new commits? (Git does not recognize that you merged them. Did you do`git checkout develop; git merge config-option`?)

PR #81, opened 2016-07-22
Remove argparse patch

lucc:
The patch is not necessary any more with python3.  This allows to parse the config file only after parsing the command line.

This will help with #79.

But this will change the command line interface slightly:  Formally you could specify sub command options for the default command without specifying the default command on the command line.  Now this is a command line parsing error.  That means if `default_command=list` this used to work but doesn't any more: `khard --addressbook main`.
I think that is acceptable.

### DIFF
--- a/khard/khard.py
+++ b/khard/khard.py
@@ -1167,45 +1167,6 @@ def copy_or_move_subcommand(action, vcard_list, target_address_book_list):
                     break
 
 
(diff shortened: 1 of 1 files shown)


lucc:
@scheibler As I take from your comments in #79 the new behavior is not acceptable. The new idea I mentioned there might also fix this. So this PR will also be stalled.

lucc:
Included in #79.

PR #82, opened 2016-07-24
readme: update Baïkal url

mathstuf:


mathstuf:
Hmm. Nevermind, the original URL is now working.

PR #83, opened 2016-07-24
setup.py: decode the readme explicitly

mathstuf:
If the LANG environment doesn't use UTF-8, we need to tell Python that
it isn't whatever it says it is explicitly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
     author_email = 'email@eric-scheibler.de',
     url = 'https://github.com/scheibler/khard/',
     description = 'A console carddav client',
-    long_description = open('README.md').read(),
+    long_description = open('README.md', 'rb').read().decode('utf-8'),
     license = 'GPL',
     keywords = 'Carddav console addressbook',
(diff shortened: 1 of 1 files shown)


scheibler:
Is this still necessary on python3?

mathstuf:
Yes, if `LANG=C`, Python goes dumb again :( .

scheibler:
Actually PyPI wants the long description in restructured text format. So I've made some additional
changes to setup.py of develop branch (convert to rst, if pypandoc module is installed, else default
to markdown).  Please, could you checkout branch develop again and verify, that your change is still
working?

mathstuf:
It needs to be `rb` for the open mode so it doesn't try to decode anything.

scheibler:
Ok, fixed.

I thought, that

```
s = open('README.md', 'r', encoding='utf-8').read()
```

and

```
s = open('README.md', 'rb').read().decode("utf-8")
```

would produce the same result, because in the first command, the encoding is specified explicitly. But
anyway, it's working for me now.

PR #85, opened 2016-07-30
setup.py: Specify encoding for open()

encukou:
This should make setup.py work in environments where UTF-8 is not the default locale.


encukou:
Never mind, I didn't notice the develop branch. Sorry for the noise!

PR #91, opened 2016-09-01
Turn last print statements into function calls

lucc:
This is one step to make all the other scripts also py3 compatible.


lucc:
was merged but github doesn't get it.

PR #92, opened 2016-09-01
Make some parts of the code more object oriented and pythonic

lucc:
The first few commits remove simple getter functions that are not needed in python and are also not considered pythonic in style (it is more a Java style).

Especially 82964c5 is moving some code into the correct class, because in object oriented programming one class should know as little as possible about the internal details of another class.

If you like these changes I can try to add more commits.  But it is always best if you merge them soon as this reduces the chances of merge conflicts that are tedious to resolve.

### DIFF
--- a/khard/actions.py
+++ b/khard/actions.py
@@ -3,6 +3,9 @@
 
 class Actions:
 
+    """A class to manage the names and aliases of the command line
+    subcommands."""
+
     action_map = {
         "add-email":    [],
         "addressbooks": ["abooks"],
@@ -44,7 +47,7 @@ class Actions:
 
         :param action: the action name to look up
         :type action: str
-        :returns: the list of aliases or None
+        :returns: the list of aliases corresponding to the action or None
         :rtype: list(str) or NoneType
 
         """
--- a/khard/address_book.py
+++ b/khard/address_book.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+"""A simple class to load and manage the vcard files from disk."""
 
 import glob
 import logging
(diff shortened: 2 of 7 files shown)


scheibler:
In principle I a appreciate your work but bigger code changes like this or the changes in the config
file require relatively much reviewing time and at the moment I can't afford that. It's already
hard to keep up with the daily business like for example the new vobject parsing errors. So I won't
drop your changes but it could take some time for merging unless it's something really small like
the added aliases pr. Thank you.

lucc:
Most of these commits can be cherry picked on their own. So if you want to just spend little time, you can have a look at one or some commits and cherry pick them. If I notice that I will rebase the rest.

All the "remove..." commits are extremely small for example. the docstring and exception commits as well.

lucc:
If it makes this easier for you I can also split this into several PRs.

lucc:
@scheibler are you interested in merging some of these? Should I open smaller PRs for some stuff? I am interested in doing some stuff again but many of the things here are in my opinion important cleanups before I want to concentrate on more convoluted changes (for example to help improve #103).

lucc:
Rebased to fix conflict after 9c12f1864c8afc8d8de4316d86c8f0823c5dc063.

lucc:
Rebased again after 0.11.4.

scheibler:
Pushed into develop.
Nice changes, thank you.
Performance was already improved (0.34s instead of 0.41s for ls of 80 contacts).

lucc:
Thank you! (I intended these changes to only tidy up the code a bit in order to prepare for other changes. But it's nice if it already improved the performence.)

PR #93, opened 2016-09-02
Add aliases to zsh completion

lucc:


lucc:
Was merged but github doesn't get it.

PR #98, opened 2016-09-28
In zsh completion, use _guard so search terms don't block option completion.

okapia:
Also, fixes to follow normal zsh conventions of lowercase singular form descriptions. That ends up being most of the patch,

### DIFF
--- a/misc/zsh/_khard
+++ b/misc/zsh/_khard
@@ -40,7 +40,7 @@ _arguments -C -s \
   '--debug[enable debug output]' \
   '--skip-unparsable[skip unparsable vcard files]' \
   ':subcommand:->subcommand' \
-  '*::options:->options' && ret=0
+  '*::option:->options' && ret=0
 
 case $state in
   subcommand)
@@ -64,7 +64,7 @@ case $state in
       {source,src}:'edit the source vcard of a contact'
     )
     # Use this array to complete the subcommands.
-    _describe -t subcommands 'khard subcommands' subcommands_array && ret=0
+    _describe -t subcommands 'khard subcommand' subcommands_array && ret=0
     ;;
 
   options)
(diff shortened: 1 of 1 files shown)


misc/zsh/_khard
```diff
@@ -88,35 +88,35 @@ case $state in
...
     )
     # search options
     local -a default_search_options merge_search_options
     default_search_options=(
-      '(-c)'{-c,--search-in-source-files}'[Look into source vcf files to speed up search queries in large address books]'
+      '(-c)'{-c,--search-in-source-files}'[look into source vcf files to speed up search queries in large address books]'
       '(-e)'{-e,--strict-search}'[narrow contact search to name field]'
       '(-u)'{-u+,--uid=}'[select contact by uid]:uid'
-      '*::search terms'
+      '*:: :_guard "^-*" "search term"'
```
lucc: I would keep the description in plural form as the user can still give more than one argument as search terms.

lucc:
From the title I would think that `khard ls foo -` and then pressing TAB should complete options. Bur for me it only gives the message ```No matches for: `option'```. Could this be an artifact of my configuration or do you get the same?

misc/zsh/_khard
okapia: Is up to you but as I indicated, the zsh convention here is to use singular. The rationale is that only one search time can be specified at a time even if it is one of a list. Completion only deals with the part being entered now, not what comes next.

okapia:
That's separate issue and I've pushed an update to fix it. _arguments was passed `-A "-*"` which indicates that no more options are valid after the first word that doesn't match the pattern. For khard, _arguments' -S option does seem to be appropriate, however. That means that if `--` appears on the line, then options are not valid after it.

_guard fixes  the case for `khard ls --sear<tab>`. As it was, it would prompt for both 'search term' and 'option' and not actually fill in the rest of --search-in-source-files because for all it knew, --sear could be part of a search term. The pattern specified with _guard tells it not to bother offering 'search term' if what is on the command-line matches the pattern.

okapia:
It's your project and it is up to you but if you're going to close an pull request unmerged that someone has taken the time to prepare then it would be nice to be afforded the courtesy of at least a brief explanation.

There's only the comment from lucc pointing out a completely separate issue (and not one that my change introduced) and I addressed that with an additional push.

scheibler:
My workflow seems a bit unconventional. I open a new local branch, fetch the pull request, review
the changes locally, merge it into develop and then close the pr. So I can avoid the github website.
But I have to remember, that I must attach a short message upon closing. Otherwise you don't know,
what happened with your pr. I'am sorry for that.

I was happy with your changes and merged them. It's commit c77106df999c50a9054e2ef9b9ba609b0d57a316.

Thank you.

lucc:
@scheibler how does your workflow look exactly (in terms of commands) because I suspect there is not only a `git merge` in there but also a `git rebase` somewhere. This changes the commits and Github will not recognice them as the commits from the PR any longer.

I guess your `.git/config` has this part:
```
[remote "origin"]
	url = https://github.com/scheibler/khard
	fetch = +refs/heads/*:refs/remotes/origin/*
```
you can change it to
```
[remote "origin"]
	url = https://github.com/scheibler/khard
	fetch = +refs/heads/*:refs/remotes/origin/*
	fetch = +refs/pull/*:refs/remotes/origin/pr/*
```
and the next `git fetch --all` or so will create a lot of branches of the form `origin/pr/987/{head,merge}`. Then you can review however you like and if you `git checkout develop && git merge origin/pr/987/head` in the end Github should probably mark the PR as "merged" insead of "closed". 

(Anyway how do you close the PRs if you don't visit the website?)

okapia:
Ok. Thanks for clarifying that. And for merging it.

PR #118, opened 2017-02-22
 Apply ISO 8601

ploth:
Change date and time representation of the birthday to an
international standard

### DIFF
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -750,9 +750,10 @@ class CarddavObject:
                 return "--%.2d-%.2d" % (date.month, date.day)
             elif (date.tzname() and date.tzname()[3:]) or \
                     (date.hour != 0 or date.minute != 0 or date.second != 0):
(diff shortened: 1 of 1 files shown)


scheibler:
The get_formatted_birthday function should return the birthdays in a human readable form. Therefore
I explicitly didn't use the international standard format.

locale.D_FMT should return a date format string for your used locale (see
        https://docs.python.org/2/library/locale.html#locale.D_FMT).

Do you have problems with the date formatting? Please provide further details.

ploth:
I am explicitly against a formatting depending on the locale.
My locale seems to be incorrect since khard displayed me a birthday like this: MM/DD/YYYY
I am also interested in the fact where I can set the locale correctly but also think we will live in better world when everyone just uses ISO 8601.
[Relevant xkcd](https://xkcd.com/1179/)

scheibler:
I've added the new config parameter localize_dates. Please pull the latest code from the develop
branch and put the lines below into your khard config file to get the date iso-formatted again:
    
```
[contact table]
...
localize_dates = no
```

PR #119, opened 2017-02-23
Some minor fixes

lucc:


scheibler:
Merged into develop with two minor fixes.

For future changes: When should we use print and when logging? For
example take the exit function on top of config.py. That also could
change to logging.error. On the other hand we could change back my
logging.error to an ordinary print statement. Any suggestions how to
proceed here?

lucc:
When I decide to use `print()` or `logging.*()` I try to figure out if the code in question is part of the terminal user interface that means if it needs user interaction or if it could also be part of a library for a gui application. If the code fits into the user interface category I use `print()` if it fits into the library category I use `logging`.

lucc:
Related to the above distinction I also think we can exchange many calles to sys.exit with raising (custom?) exceptions. Then we could resort to call sys.exit only from the main module. So we can regard some of our modules to have more a status of a library and some more a status of a user interface.

PR #122, opened 2017-02-26
Misc simplifications

lucc:
These are some misc simplifications I found along the way (while trying to speed up khard in different ways, experiments ongoing).


lucc:
@scheibler are you interested in this, should I update it?

lucc:
Updated to latest master, will merge if nobody vetos.

PR #123, opened 2017-02-26
Misc fixes

lucc:
These are some fixes for small bugs (typos and such) that I found while doing other stuff.


lucc:
Updated to latest master, will merge if nobody vetos.

PR #125, opened 2017-03-05
More object oriented address book

lucc:
This PR moves as much responsibility about managing collections of vcards to the address book (as I believe it is not really the responsibility of the config class to manage vcards).

Also the AddressBook class is split up so it can be extended more easily.  Currently an abstract base class, one class to load from a directory of vcard files and one class to collect many such folders into one abstract address book are implemented.  The AddressBookCollection class takes over many of the responsibilities of the config class about managing vcards from different address books.  One idea to extend this could be that we could use a cache to reduce the time needed for parsing all the vcards as it looks like this is the main bottle neck for khard's performance at the moment.  But that is a discussion for another PR/issue.

I can notice a very small performance improvement with this PR.  I use this zsh code to test it:
```zsh
export PYTHONPATH=.
upsteam=develop
count=10
head=$(git rev-parse --abbrev-ref HEAD)
time (for i in {0..$count}; do ./khard-runner.py ls &>/dev/null; done )
git checkout $upstream
time (for i in {0..$count}; do ./khard-runner.py ls &>/dev/null; done )
git checkout $head
```
I assume that this improvement is due to my reimplementation of the `get_short_uid_dict()` method which I did clean up and free of some needless recomputations of the same constant values.

### DIFF
--- a/khard/address_book.py
+++ b/khard/address_book.py
@@ -278,7 +278,7 @@ class VdirAddressBook(AddressBook):
         :returns: the number of successfully loaded cards and the number of
             errors
         :rtype: int, int
-
+        :throws: AddressBookParseError
         """
         if self.loaded:
             return len(self.contacts), 0
--- a/khard/address_book.py
+++ b/khard/address_book.py
@@ -21,22 +21,16 @@ class AddressBook(metaclass=abc.ABCMeta):
 
     """The base class of all address book implementations."""
 
-    def __init__(self, name, path):
+    def __init__(self, name):
         """
         :param name: the name to identify the address book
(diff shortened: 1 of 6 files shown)


josephholsten:
@lucc this looks interesting, but can you rebase it onto master?

lucc:
sorry @josephholsten it took so long that the rebase is getting complicated. I will only do it if @scheibler is still intrested in any of this. From his recent reactions (none) it seems not so much. (@scheibler?)

lucc:
@josephholsten I rebased it on the latest develop branch.

Issue #148, opened 2018-01-12
Contact Table config option is now required


pigmonkey:
I just upgraded to the newly released v0.12.0. Upon running khard, I received an error about a missing config option:

```
$ khard 
Traceback (most recent call last):
  File "/home/pigmonkey/.virtualenvs/khardtest/bin/khard", line 11, in <module>
    sys.exit(main())
  File "/home/pigmonkey/.virtualenvs/khardtest/lib/python3.6/site-packages/khard/khard.py", line 1478, in main
    args = parse_args()
  File "/home/pigmonkey/.virtualenvs/khardtest/lib/python3.6/site-packages/khard/khard.py", line 1454, in parse_args
    config = Config(args.config)
  File "/home/pigmonkey/.virtualenvs/khardtest/lib/python3.6/site-packages/khard/config.py", line 121, in __init__
    self.config['contact table']['sort']
  File "/home/pigmonkey/.virtualenvs/khardtest/lib/python3.6/site-packages/configobj.py", line 554, in __getitem__
    val = dict.__getitem__(self, key)
KeyError: 'sort'
```

I've never had a `contact table` in my khard configuration before. Adding this section and defining a `sort` value as demonstrated in the example config fixes the issue, but I'm not sure if this behaviour is intentional. There was nothing in the changelog about this configuration now being required. It seems like it might be better for khard to fallback to a default sort value if the user has not expressed a preference.

ilovezfs:
@scheibler  I hit the same error in the `test do` block of the Homebrew formula while trying to update it.

lucc:
I would argue that this is a "bug". If the key is intentionally required (I don't know) then one shouldn't see an exception but an error message instead.

More generally I think we can do better by providing more default values that should kick in if a user doesn't have the respective config options set (maybe it even makes sense to run khard without a config file?). This could reduce friction for first time users. If we have a man page (#104) it would also be simple for users to discover these defaults. The help output could also mention these defaults (maybe even adjusted to the values read from the config file).

lucc:
Does #155 solve this for you?

ilovezfs:
@lucc works for me.

lucc:
Fixed in 0.12.1.

PR #155, opened 2018-01-14
Fixes issue #148

scheibler:


lucc:
Looks good to me. The test failure is understanable as the travis config file was merged only in develop. We will see correct test runs on travis only after 0.13, but I think that is ok.

I quickly tested it locally and it looks like it does the job. So from my point of view this can be merged.

Issue #159, opened 2018-01-16
Buggy behaviour with `khal ls -u ...`


lucc:
I found this when working on #158 but verified the behaviour is also in
master (and develop).

I looked up the *exact* uid of a contact:
```
$ khal detail lucas hoffmann | grep UID
    UID: bsam6a3srto0wo8d8kyupdgtuuzk1igo0s8z
```
Now I look for this ID *exactly* I get this unexpected reply:
```
$ khard ls -u bsam6a3srto0wo8d8kyupdgtuuzk1igo0s8z
Found multiple contacts for uid bsam6a3srto0wo8d8kyupdgtuuzk1igo0s8z
    Lucas Hoffmann: bsam6a3srto0wo8d8kyupdgtuuzk1igo0s8z
    Lucas Hoffmann: bsam6a3srto0wo8d8kyupdgtuuzk1igo0s8z
```
When I take any initial part of the uid I get the contact just fine:
```
$ khard ls -u bsam6a3srto0wo8d8kyupdgtuuzk1igo0s8
Address book: main
Index    Name               Phone                   E-Mail                UID
1        Hoffmann, Lucas    xxxxxxxxxxxxxxxxxxxx    xxxxxxxxxxxxxxxxxx    b
```
and it gets better:  When I try to search in another address book I get the
same result!
```
$ khard ls -a other -u bsam6a3srto0wo8d8kyupdgtuuzk1igo0s8
Address book: main
Index    Name               Phone                   E-Mail                UID
1        Hoffmann, Lucas    xxxxxxxxxxxxxxxxxxxx    xxxxxxxxxxxxxxxxxx    b
```

I found a way to fix this in #158 but am still working on a backport to
master and develop.

Now I started thinking: we have `search_in_source_files` and we do shorten
uids.  But if I run a query with a small result set I also get very short
"shortened uids" this means only the files really loaded from disk are
respected for this.  And this is what `search_in_source_files` should really
do so it is ok.  But on the othre hand the short "UID" might not actually be
unique if I use it with another (broader) query.  So what would be possible
solutions to this dilemma?  Leave it as is, not shorten uids, forbid to
shorten uids if not all address books or contacts are involved?

lucc:
This should now be fixed in master and develop.

PR #160, opened 2018-01-17
Fixes a problem with the _search_all function in address_book.py

scheibler:


lucc:
I can not reproduce this. For me `khard ls eluc` returns nothing. But the code looks good and doesn't seem to break anything. (Dumb question did you check that Ricarda doesn't have the string "eric" anywhere else in the vcard?)

scheibler:
@lucc:
> did you check that Ricarda doesn't have the string "eric" anywhere else in the vcard?)

That's the point. "Ricarda" itself doesn't contain the search query "eric". But ...

First in the _search_all function, the search query regexp is matched against the output of
print_vcard() (not only the name). So we match:

"Name: Ricarda\n[...]" contains "eric" == False

That's correct.

Then look at the phone number hack in the next line where the output of print_vcard() is stripped of
all non-letters and -numbers to find phone numbers too. So now we match:

"NameRicarda\n[...]" contains "eric" == True

And that's not right.

Problem: "Nam<eRic>arda\n[...]" and the case insensitive regexp

Of course that's a really rare scenario but nevertheless not impossible. Otherwise I wouldn't
encountered it :-).

lucc:
PS: travis will fail until we merge in the stuff from develop with v0.13 so from my point of view this can be merged (if you want to cut down on releases you could merge this and #159 and release 0.12.2 only after both are merged).

scheibler:
Merged into master and develop

PR #169, opened 2018-02-02
Split and restructure carddav classs

lucc:
This series splits the old CarddavOBject class into two classes.  Most of the old class is left in place and only some code is moved to a parent class.  Some tests are also added.  The objective here is twofold:

* make the code more testable (the parent class is designed to be more testable)
* separate the knowledge about address books, our yaml format the filesystem and so on from the handling of vcard stuff (the new class wraps a vcard, the child class handles the rest)
* structure the code according to responsibility (maybe we can add more classes here, one to wrap the vcard, one to handle the connection to an address book, one to handle reading from disk, one for searching vcards)


lucc:
@scheibler: this is now in a state where it could be merged. I would like to have a short review from you if your time permits. Better even if you could use it for a bit.

lucc:
PS: reviews and test usage by other people is also helpful of course :)

lucc:
@scheibler I would like to merge this. If you don't object I can also do it myself.

(I just rebased on the current develop to solve a conflict)

scheibler:
I want to create a new release in the next two weeks cause we hadn't one in a very long time and
there are a couple of new features and bug fixes which should move into master at last.

Therefore I would suggest that we wait with rather big code changes until that release is done.
Afterwards feel free to merge.

lucc:
thanks for the heads up

PR #170, opened 2018-02-07
Update README.md

AvdN:
minor typo


lucc:
Thanks @AvdN .

@scheibler I think this does not need a new release so I merged it into master but I think we can wait for the next *real* bug fix until you have to take action again :)

AvdN:
Indeed no release necessary for such a minor thing. 
I was just checking up on khard after helping the Gentoo people to get ruamel.yaml installed for it.

PR #192, opened 2018-10-13
README: fix mutt's config sample

XeR:
Remove single quotes from the query_command in the mutt config sample to fix
code injection vulnerability

mutt already escapes arguments (%s) with single quotes.
mutt's manual states the following:
> This specifies the command Mutt will use to make external address queries.  The string may contain a
> “%s”,  which  will be substituted with the query string the user types.  Mutt will add quotes around
> the string substituted for “%s” automatically according to shell quoting rules, so you should  avoid
> adding your own.  If no “%s” is found in the string, Mutt will append the user's query to the end of
> the string.  See “query” for more information.

When `query_command` contains `'%s'`, the quotes are doubled by mutt.
`khard email --parsable '%s'` becomes `khard email --parsable ''Dave Null''` and is parsed by the shell as `["khard", "email", "parsable", "Dave", "Null"]`

While this may cause troubles with Irish names (O'Foobar), it also makes it possible to run arbitrary command from mutt's `From` field, by using special characters.
Auto-completing `; bash;` will result in mutt spawning a bash shell.
```
mutt
└─ sh -c khard email --parsable ''; bash; ''
   └─ bash
```

The exploitation of this vulnerability by an attacker is very unlikely.
Note that it is also possible to fix this by removing %s altogether.

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,7 +324,7 @@ Khard may be used as an external address book for the email client mutt. To acco
 following to your mutt config file (mostly ~/.mutt/muttrc):
 
 ```
-set query_command = "xargs khard email --parsable -- <<< %s"
+set query_command = "echo %s | xargs khard email --parsable --"
 bind editor <Tab> complete-query
 bind editor ^T    complete
 ```
@@ -334,7 +334,7 @@ address books contain hundreds or even thousands of contacts and the query proce
 may try the --search-in-source-files option to speed up the search:
 
 ```
-set query_command = "xargs khard email --parsable --search-in-source-files -- <<< %s"
+set query_command = "echo %s | xargs khard email --parsable --search-in-source-files --"
 ```
 
 To add email addresses to khard's address book, you may also add the following lines to your muttrc file:
--- a/README.md
+++ b/README.md
@@ -324,7 +324,7 @@ Khard may be used as an external address book for the email client mutt. To acco
 following to your mutt config file (mostly ~/.mutt/muttrc):
 
 ```
-set query_command = "xargs khard email --parsable <<< %s"
+set query_command = "xargs khard email --parsable -- <<< %s"
 bind editor <Tab> complete-query
 bind editor ^T    complete
 ```
@@ -334,7 +334,7 @@ address books contain hundreds or even thousands of contacts and the query proce
 may try the --search-in-source-files option to speed up the search:
 
 ```
-set query_command = "xargs khard email --parsable --search-in-source-files <<< %s"
+set query_command = "xargs khard email --parsable --search-in-source-files -- <<< %s"
 ```
 
 To add email addresses to khard's address book, you may also add the following lines to your muttrc file:
--- a/README.md
+++ b/README.md
@@ -324,7 +324,7 @@ Khard may be used as an external address book for the email client mutt. To acco
 following to your mutt config file (mostly ~/.mutt/muttrc):
 
 ```
-set query_command= "khard email --parsable %s"
+set query_command = "xargs khard email --parsable <<< %s"
 bind editor <Tab> complete-query
 bind editor ^T    complete
 ```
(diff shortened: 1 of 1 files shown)


scheibler:
These quotes hound me ... Changed them at least twice in khards lifetime.

There is a reason why I added the single quotes around %s. Let's assume the contact john contains the following two email addresses:

1. john@example.com
2. john@example.org

If I then enter "john" into mutts to field I get both email addresses and can pick one with the arrow keys. But often I want to select a specific address directly. With the single quotes around %s I simply can type "john org" into the to field and get only "john@example.org". If I remove the single quotes that doesn't work any more.

Maybe that works cause mutt starts a shell in the background. Question is how to avoid the code injection issue and preserve the described functionality?

Just for reference, I use:
- Khard version 0.12.2
- NeoMutt 1.7.2

lucc:
That could be fixed by this wrapper script:
```bash
#!/bin/bash
read -ra args <<<"$*"
khard email --parsable "${args[@]}"
```
The problem is that khard treats `khard email a b` differently from `khard email 'a b'`.  We join individual args with `.*` for the regex search but do not replace spaces inside each arg.

XeR:
> Question is how to avoid the code injection issue and preserve the described functionality?

Oh, I did not know about this functionality. (I'm a new khard user!)
@lucc's wrapper script would work, but I am not fond of using an external script.
My workaround uses `xargs` and does not require dropping an additional file.

I also noticed that the previous fix was incomplete: I missed one of the two mutt config samples. On top of that it was still possible to inject parameters (i.e. -h, or -v)

This should hopefully make everything work as intended!

I like to have my commits as light as possible (mostly for git revert)
Should I merge the commits in a single commit, or should I keep them that way?

README.md
```diff
@@ -324,7 +324,7 @@ Khard may be used as an external address book for the email client mutt. To acco
 following to your mutt config file (mostly ~/.mutt/muttrc):
 
 ```
-set query_command= "khard email --parsable %s"
+set query_command = "xargs khard email --parsable <<< %s"
```
lucc: The `<<<` is `bash` syntax and will not work in `sh` or `dash`!

lucc: The `<<<` is `bash` syntax and will not work in `sh` or `dash`!

README.md
```diff
@@ -324,7 +324,7 @@ Khard may be used as an external address book for the email client mutt. To acco
 following to your mutt config file (mostly ~/.mutt/muttrc):
 
 ```
-set query_command= "khard email --parsable '%s'"
+set query_command = "xargs khard email --parsable -- <<< %s"
```
lucc: The `<<<` is `bash` syntax and will not work in `sh` or `dash`!

XeR:
Nice catch. My distro symlinks sh to bash, so I did not realize that.
What about `echo | xargs`? The only problem I can think of is when searching for `-n`

lucc:
I thought about this again and I figure there are two issues here:
1. We should not put the quotes in the example usage.
2. The use case @scheibler describes is unrelated to point one and should have an independent fix. I suggest a new option `--split-at-whitespace` or similar.

Additionally to point one I vote that we also not make the example usage with mutt to complicated. There is no real problem with `khard email --parsable %s` as the query command. Only the inconvenience of point two.

After the new release that @scheibler currently prepares I will have a look at #169 and this one afterwards. I think you can clean up this PR into a single commit and just fix the issue with the query command itself. No hacks for special search behaviour with whitespace.

scheibler:
I removed the '' around %s in the readme file cause I agree that this command should be clear and
simple and that we should fix that in khard directly in the future.

But I also tried the provided "xargs" solution and found it quite suitable. Therefore I've included
that as well as an alternative for those, who want to search for more than a single word at once.

Thank you for sharing.

Issue #179, opened 2018-05-06
Redesigning the command line interface


lucc:
I have collected some thoughts and ideas about the command line interface of khard and how it could be improved. I will share them here and hope for people joining the discussion. 

# status quo

Currently we have 16 subcommands and some aliases. I would order them into these categories:
- listing several contacts: `ls`, `bdays`, `email`, `phone`, `file`
- displaying info about one contacts only: `show`, `export`
- modify contents of a single contact: `edit`, `source`
- some misc commands: `merge`, `copy`, `move`, `rm`, `abooks`, `new`, `add-email`

# plan

I think the commands for the first three categories can be unified into three general commands with some new options (first the easy ones :)

- modify: we could have one single command `edit` and it takes one new option `--source` (and `--yaml` or just `--format={source,yaml}`) to provide the behaviour of the current `source` command.
- same for info about one contact: only keep the `show` command and introduce an option `--format={pretty,yaml,vcard}` to provide the current `export` functionality (and even a new one with `vcard`: just print the contents of the vcard on stdout)
- also merge all listing commands into one.  This is a little harder as I am not sure how we can put all these different formats into a new `--format` option.

All old command names could be kept as a kind of compatibility alias for some time.

## format option for new `ls`

For compatibility the `--format` option could accept some special strings to mimic the exact behaviour of the old listing commands. 

I light of some open pull requests about special/new listing options for some subcommands (#175, #89) we could also accept python format strings that would be interpolated with the contact objects.  Then people could write their own format strings.  This would also open up khard for use in custom shell scripts (which is a thing that I also would like).

One problem with format strings is that I do not know how we can implement alignment over the whole printed list. We currently do that by generating a python `list` with strings for each entry and then search for the longest entry in each column to adjust the amount of whitespace.  With plain format strings this is not possible.  One would have to implement a special string at which to inject alignment (but that introduces string parsing and escaping issues into the new format strings) or we would have to support lists of format strings so that we can insert the alignment between them. Lastly we could ignore the alignment issue for custom, user written format strings and only support it in our buildin special format name (which are meant to mimic the format of our current subcommands).

Another problem with custom format strings is that some attributes of the vcard objects are lists and we would have to decide if the format string is applied to the whole list or for each entry in turn. This corresponds to the problem in #173.

If I have enough time I will open PRs for the first to command unifications. For the listing commands I will wait for some comments first.

earboxer:
[...]

Issue #131, opened 2017-04-22
Search within specific fields


GSI:
How are search terms currently considered - and Is there a way to filter by certain properties of contacts?

Something like `khard details category:2016 prefix:dr` would return any doctors in the 2016 category.

lucc:
Currently the search terms are used as a regex that is matched against the rendered vcard. But this proposal is very nice. It reminds me of https://github.com/notmuch/notmuch .

On the technical side: We have an option `--search-in-source-files` that searches for the regex in the source file (unrendered vcard) before parsing it (to save time). How should such a query be interpreted by this option?

lucc:
@GSI I like your idea even more after I thought about it for some time. So here are my thoughts:

We are currently keeping the order of search terms (so for `khard ls foo bar` we search for `foo.*bar` but not `bar.*foo`) but that does not make sense for searching in fields because they don't have any inherent order.  So in general one could relax the current implementation by using the python `in` operator on strings instead of regex searches and combine that with `all()`. This would effectively remove the order from search terms.

From there it might be easier to implement searching in specific fields. I imagine that search terms that do not start with a known field name are treated as before. This means that `khard ls name:foo` should look for "foo" in the name, `khard ls foo` should look for "foo" anywhere in the cards and `khard ls foo:bar` should look for "foo:bar" in any field.

One further question is how multible search terms for one field should be treated. So what does `khard ls name:foo name:bar` search for? For "foo" *and* "bar" in the name or for any of "foo" *or* "bar"? Or is it an error? We could also solve this question by introducing two new command line options `--all` and `--any` which could use the python functions of the same name to implement both of the above mentioned search strategies.

I am currently refactoring the basic class that represents a vcard in #169. I think that can help us to achieve the above points. You are welcome to have a look at it and leave your comments.

GSI:
Quick comment due to time constraints:

Very reasonable thoughts! I'd "and" anything together. For "or" one may aways run two separate searches in sequence.

vleonbonnet:
I created #237 to add a category search. This does not cover all fields, but covers my own workflow and be of use to others until a full revamp of the search feature is completed.

earboxer:
I think for this, it would be better to output the vcards as json objects, and have the users who want this level of filtering use the program `jq`.

`khard ls --json | jq '. | map(select((.titles[0] == 'dr' or .first_name == 'doctor') and '2016' in .category))` or similar.

lucc:
[...]

PR #199, opened 2018-12-27
Redesign command line interface

lucc:
This is the corresponding PR for #179.  Until now the unification of show-export and edit-source is done. The unification of the different ls versions is underway.

I am happy about any feedback.

### DIFF
--- a/khard/actions.py
+++ b/khard/actions.py
@@ -11,7 +11,6 @@ class Actions:
         "addressbooks": ["abooks"],
         "birthdays":    ["bdays"],
         "copy":         ["cp"],
-        "details":      ["show"],
         "email":        [],
         "export":       [],
         "filename":     ["file"],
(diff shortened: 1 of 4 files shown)


lucc:
There is one problem with the unification of the different ls subcommands. At least the phone subcommand tries to match the search terms against the phone numbers and not the rest of the card.  This makes sense but is harder to unify with the other listing like subcommands. I think I will try to rework the matching of cards first to have a better documented api for matching cards in specific ways. This could also help with implementing something like #131.

In conclusion I will leave this PR as is and only add more code in a new PR.

lucc:
@scheibler as this feels like a fundamental decision to me and is also user facing I would like to have a short ok/no from you before I proceed.

lucc:
@scheibler  ping?

legrostdg:
@scheibler ping?

Issue #104, opened 2016-12-02
add a manpage


legrostdg:
It would be cool to add a manpage to khard.

Unip0rn:
As well as for the config file, too.

lucc:
I will hijack this thread and extend the question a litte bit: We should restructure the way khard's user documentation is organized. Currently we have the help output and the README file. To me this division would make sense:
1. the README: should be smaller and just present the main features and idea of khard+instructions on installing and contributing (could be moved to a dedicated CONTRIBUTING file later)
2. the man page(s?): should contain most of the in depth documentation currently found in the README. If the examples and tips grow out of bounds they could be moved to the wiki.
3. the help output: is mostly good as it is right now, only that help output for subcomands doesn't work without a config file.

lucc:
We could use sphinx to build all kinds of documentation.

lucc:
We could also use github pages to serve the docs as our website: [official docs](https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/#publishing-your-github-pages-site-from-a-docs-folder-on-your-master-branch)

I think we can even use the sphinx docs for that. It looks like [khal](https://github.com/pimutils/khal) does that.

jelmer:
It would be nice to have a manpage (regardless of other documentation); some distributions, e.g. Debian require a manpage for every binary.

lucc:
@legrostdg I merged a man page in #201. I hope that covers the basics of this feature request. For further improvements to the documentation I opened #204. I invite everybody interested to join the discussion there.

legrostdg:
It's great, thanks!

PR #201, opened 2019-01-01
Add man page and other docs with sphinx

lucc:
This adds a docs folder with a man page and a general help document.  Some of the contents of the document are taken from the readme. This is done to make the readme smaller and easier to check for new visitors. Only users who want to try khard should be bothered with the in depth documentation.

The man page was also requested in #104. I think this might be a good idea as we are already present in quite some repositories: https://repology.org/metapackage/khard/versions

@scheibler I would like to have a short comment from you about this plan if your time permits:
1. Add a man page (done here)
2. Add sphinx docs (done here)
3. Move most real documentation from the readme to the sphinx docs. The readme should only be the entry point for new visitors on github and pointers for people who downloaded the source distribution (tarball).
4. With the structure given by these points we can extend the docs as much as we want without bloating the readme file.
5. look into serving the docs as webpage (probably with GH pages)


scheibler:
Yes, very good start. Thank you.

- A man page is definitely required by now.
- The Readme should at least contain the pip installation instructions and a few usage examples.

lucc:
Thanks for your feedback. I will work on this topic some more this week.

PR #207, opened 2019-03-09
Display ABLABELs for URLs and Private Objects

d7415:
Also enable adding/editing these.

Recreated #202 because git wouldn't accept my merging onto master and back onto develop and github won't let me change source to develop (or my feature branch).

The history is a big mess, but it should apply cleanly. I'm not sure at this point how to make the PR cleaner without repeating a lot of work. I'm open to suggestions though...


d7415:
No, I'll try again... That's not manageable for reviewing.

PR #208, opened 2019-03-09
Display ABLABELs for URLs and Private Objects

d7415:
Also enable adding/editing these.

Recreated #202 because the merge onto develop wasn't possible without (someone else's) manual intervention.

I still need to add tests.

### DIFF
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -57,7 +57,6 @@ def convert_to_vcard(name, value, allowed_object_type):
     raise ValueError("Error: " + name +
                      " must be a string or a list with strings.")
 
-
 class VCardWrapper:
     """Wrapper class around a vobject.vCard object.
 
@@ -133,7 +132,10 @@ class VCardWrapper:
         for child in self.vcard.getChildren():
             if child.name == name:
                 ablabel = self._get_ablabel(child)
-                values.append(ablabel + (": " if ablabel else "") + child.value)
+                if ablabel:
+                    values.append(ablabel + ": " + child.value)
+                else:
+                    values.append(child.value)
         return sorted(values)
 
     def _delete_vcard_object(self, name):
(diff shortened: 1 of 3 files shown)


d7415:
Ok. I think I've now addressed everything.

I'm not entirely happy with the yaml unit test - The input doesn't have the ':' escaped in the URL because if I do, the backslash is escaped. I guess that stuff is handled before `CarddavObject.from_user_input` gets hold of it.

test/test_yaml.py
```diff
@@ -73,3 +73,28 @@ def test_empty_note_in_yaml_input(self):
...
+        """
+        # Careful, this function doesn't actually support named arguments so
+        # they have to be kept in this order!
+        return CarddavObject.from_user_input(
+                address_book=mock.Mock(path='foo-path'), user_input=yaml,
+                supported_private_objects=[], version='3.0',
+                localize_dates=False)
+
+    def test_ablabelled_url_in_yaml_input(self):
+        ablabel_url = "First name: foo\nWebpage:\n - http://example.com\n" +\
```
lucc: You can use quotes in YAML to disambiguate a string from a dict, so 
```yaml
Webpage:
 - "http://example.com"
 - github: "https://github.com/scheibler/khard"
```
should work.  But I though YAML only uses colons without a following space for dicts. So is there actually a syntax problem in your YAML or not?

And you don't need the `+` here as Python has implicit concatenation of string *literals*. That means `"foo"  "bar" == "foobar"`

khard/carddav_object.py
```diff
@@ -57,7 +57,6 @@ def convert_to_vcard(name, value, allowed_object_type):
     raise ValueError("Error: " + name +
                      " must be a string or a list with strings.")
 
-
```
lucc: [PEP8](https://www.python.org/dev/peps/pep-0008/) tells us to use two empty lines between classes.

khard/carddav_object.py
```diff
@@ -333,6 +339,71 @@ def anniversary(self):
...
+            else:
+                return group_name
+
+    def _add_labelled_object(self, obj_type, user_input, name_groups=False):
+        obj = self.vcard.add(obj_type)
+        if isinstance(user_input, dict):
+            if len(user_input) > 1:
+                raise ValueError("Error: %s must be a string or a dict " +\
+                                 "containing one key/value pair." % obj_type)
+            label = [i for i in user_input][0]
```
lucc: Sorry it is `list(user_input)[0]` not `user_input.keys()[0]` as I previously suggested. The list comprehension `[x for x in y]` just looks like a noop so `list(y)` might be more comprehensible.

test/test_yaml.py
d7415: > You can use quotes in YAML to disambiguate a string from a dict, so
> 
> should work. But I though YAML only uses colons without a following space for dicts. So is there actually a syntax problem in your YAML or not?

I was trying to make it reflect the behaviour of `khard edit`, which gives the URL as `https\://`, possibly to avoid the unlabelled URL becoming a dict. It certainly interprets correctly in-app.

The test works with or without quotes - I can add them if you want.

> And you don't need the `+` here as Python has implicit concatenation of string _literals_. That means `"foo" "bar" == "foobar"`

Ah. I didn't realise that worked across lines (with `\`)

khard/carddav_object.py
d7415: Ah, type casting :) In that case, I can use `list(user_input)[0]`

d7415:
Just that yaml test left, whether you want the quotes included or not. As I said in-line, it passes either way.

PR #210, opened 2019-04-24
Add an option to accept 0 as "None" in `choose_vcard_from_list()`. Us…

### LINKED ISSUE #209: khard mv wants to overwrite contact
I want to move a contact from one address book to another. (details changed) It looks like this:
````
> khard "something unique"        
Address book: default
Index    Name        Phone                    E-Mail    UID
1        Erika    pref: +43 ... .......              1    
> khard mv -A old "something unique"        
Move contact Michaela from address book default
Select target contact which to overwrite
Address book: old
Index    Name               Phone                           E-Mail                                                UID
1        Erika Haus    WORK, VOICE: +43 . .........    INTERNET, WORK, pref: something@something                 123       
2        Erika Berg                                    INTERNET, WORK, pref: somethingelse@somethingelse         345    
Enter Index: 0
Please enter an index value between 1 and 2 or nothing to exit.
Enter Index: 
Canceled
````
I don't want to overwrite any entry, it should just be moved to the other addressbook as new entry. But it offers me two choices to overwrite and when I don't select either, the move is aborted.

d7415:
…e this option in `copy_or_move_subcommand`.

Fixes #209.

### DIFF
--- a/khard/khard.py
+++ b/khard/khard.py
@@ -429,8 +429,9 @@ def choose_vcard_from_list(header_string, vcard_list, include_none=False):
     list_contacts(vcard_list)
     while True:
         try:
-            prompt_string = "Enter Index" + \
-                           (" (0 for None)" if include_none else "") + ": "
+            prompt_string = "Enter Index (" + \
+                            ("0 for None, " if include_none else "") + \
+                            "q to quit): "
             input_string = input(prompt_string)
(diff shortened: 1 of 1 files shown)


d7415:
There may be a better way of doing this, but this seemed to make sense.

PR #220, opened 2019-08-11
Handle case where _get_multi_property() encounters a list instead of …

d7415:
…a string.

In the ABLABEL-related code in `_get_multi_property()` it assumes the result is a string. Today I encountered one that presented itself as a list (containing a single string). This should handle strings, lists containing a single string and (slightly messily) longer lists.

I'm open to suggestions if you have a better idea.

### DIFF
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -134,7 +134,7 @@ class VCardWrapper:
             if child.name == name:
                 ablabel = self._get_ablabel(child)
                 if ablabel:
-                    values.append(ablabel + ": " + child.value)
+                    values.append(ablabel + ": " + "".join(child.value))
                 else:
                     values.append(child.value)
         return sorted(values)


lucc:
Can you supply an example vcard so I can investigate the issue as well?

d7415:
Ok, I can't create a contact to reproduce this using khard. It might be Google creatively interpreting the spec (for no clear reason in this case).

An example combination that will break khard:

```
item2.ORG:ACME Inc
item2.X-ABLabel:Work
```

However when editing in khard, it displays it (using this patched version) in the template as

```
Organisation : Work: ACME Inc
```

but if the card is changed, it shows

```
mapping values are not allowed in this context
  in "<unicode string>", line 46, column 20
```

So khard certainly doesn't think this is valid.

Google are exporting contacts in this format, and currently khard won't load at all if a contact is present, without any indication of which contact or field is the issue. This patch allows khard to read these fields, so a user can at least tell which contact (and field) has a problem if and when they encounter it.
Alternatively, I could look into expanding the PR to allow editing fields in this way, but that might not be best practice - I'm not sure.

----
Error on using khard:
```
Traceback (most recent call last):
  File "bin/khard", line 11, in <module>
    load_entry_point('khard', 'console_scripts', 'khard')()
  File "/home/martin/tmp/khard/khard/khard.py", line 1787, in main
    vcard_list = generate_contact_list(config, args)
  File "/home/martin/tmp/khard/khard/khard.py", line 712, in generate_contact_list
    args.strict_search if "strict_search" in args else False)
  File "/home/martin/tmp/khard/khard/khard.py", line 470, in get_contact_list_by_user_selection
    config.reverse(), config.group_by_addressbook(), config.sort)
  File "/home/martin/tmp/khard/khard/khard.py", line 496, in get_contacts
    contacts.extend(address_book.search(query, method=method))
  File "/home/martin/tmp/khard/khard/address_book.py", line 89, in _search_all
    contact_details = contact.print_vcard()
  File "/home/martin/tmp/khard/khard/carddav_object.py", line 1656, in print_vcard
    if self.organisations:
  File "/home/martin/tmp/khard/khard/carddav_object.py", line 579, in organisations
    return self._get_multi_property("ORG")
  File "/home/martin/tmp/khard/khard/carddav_object.py", line 137, in _get_multi_property
    values.append(ablabel + ": " + child.value)
TypeError: can only concatenate str (not "list") to str
```

Full example contact:
```
BEGIN:VCARD
VERSION:3.0
UID:51953f58-237e-429a-a271-cb47c026a895
EMAIL;TYPE=INTERNET,HOME:jsmith12345@example.net
EMAIL;TYPE=INTERNET,WORK:john.smith@example.com
FN:John Smith
N:Smith;John;;;
NICKNAME:Smithy
NOTE:Favourite Colour: Red\nLucky Number: 7
item2.ORG:ACME Inc
item2.X-ABLabel:Work
REV:201908dT114309Z
TEL;TYPE=cell:1234567890
TEL;TYPE=home:0987654321
END:VCARD
```

lucc:
Thank you for the example. I used this minimal version
```
BEGIN:VCARD
VERSION:4.0
UID:51953f58-237e-429a-a271-cb47c026a895
FN:labeled guy
item1.ORG:Test Inc;Some devision
item1.X-ABLabel:Work
END:VCARD
```
The problem with you proposal is that `ORG` can have several fields seperated with `;` (that's why it is a list). But with your simple `''.join()` for once the current behaviour of representing the organisation in a structured way in the yaml output/pretty printing is lost. Additionally the formatting looks wrong when joining with the empty string.

We will have to find some other solution that handled rich value types like `ORG` and `ADR` or `NAME` correctly and still can cope with ablabels.

I will open a new issue for this, you are welcome to post PRs if you have any ideas for a solution.

PR #222, opened 2019-08-19
Make FN editable

### LINKED ISSUE #89: Display "name" or "nickname" instead of "first name" + "last name"
I have some contacts that have the same first and last names and differ only in their middle name. When I attempt to select one from `mutt`'s menu, or the native `khard` menu, the displayed names seem to be derived from the "first name" + "last name" instead of the actual name.

For example, I have the following contact. 

```
Name: John Foo Smith
First name : John
Additional : Foo
Last name  : Smith
```

and also

```
Name: John Bar Smith
First name : John
Additional : Bar
Last name  : Smith
```

From khard's native interface, it's difficult to differentiate between the two.

```
$ khard modify smith
Select contact for Modify action
Address book: owncloud
Index    Name          Phone                 E-Mail                   UID   
1        John Smith    CELL: 1234 567 890    WORK: jbsmith@baz.com    49    
2        John Smith    CELL: 1234 567 891    WORK: jfsmith@baz.com    5a4   
Enter Index:
```

Preferably, the basic contact "name" should be displayed. Another good option would be to display the "nickname" instead, although I guess this would need to be an option since not everyone would like that. Personally, the latter seems useful to me, but the first option is probably "safer".

### LINKED ISSUE #171: Add FN (Full Name) of vCard to display configuration
Also mentioned in #89 by @madduck , there's a "FN" field in vCard.

Not only same first name and last name people exists in real world, there are also some countries put last name first and first name afterwards, i.e. China, Japan, etc.

As a result, it would be better if 'FN' is configurable in khard. Thx.

lucc:
See #171 for the initial discussion.

closes #171 
closes #89

### DIFF
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -469,8 +469,19 @@ class VCardWrapper:
         :type value: str
         """
         self._delete_vcard_object("FN")
-        self.vcard.add("FN").value = convert_to_vcard("FN", value,
-                                                      ObjectType.string)
+        if value:
+            value = convert_to_vcard("FN", value, ObjectType.string)
+        elif self._get_first_names() or self._get_last_names():
+            # autofill the FN field from the N field
+            names = [self._get_name_prefixes(),
+                     self._get_first_names(),
+                     self._get_last_names(),
+                     self._get_name_suffixes()]
+            name = [x for x in names if x]
+            value = helpers.list_to_string(names, " ")
+        else:  # add an empty FN
+            value = ""
+        self.vcard.add("FN").value = value
 
     def _get_names_part(self, part):
         """Get some part of the "N" entry in the vCard as a list
@@ -545,7 +556,6 @@ class VCardWrapper:
 
     def _add_name(self, prefix, first_name, additional_name, last_name,
                   suffix):
-        # n
         name_obj = self.vcard.add('n')
         stringlist = ObjectType.string_or_list_with_strings
         name_obj.value = vobject.vcard.Name(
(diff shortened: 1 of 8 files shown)


khard/helpers.py
```diff
@@ -202,6 +202,10 @@ def get_new_contact_template(supported_private_objects=None):
 
     # create template
     return dedent("""
+        # Every contact must contain a formatted name, it will be autofilled
+        # from the full name below if not given.
+        Formatted name :
```
hydrargyrum: Isn't it rather `full name` when it's the input, and `formatted name` when it's generated by first name/last name templating?

lucc: It represents the value in the `FN` field of the vcard.  And that is
> formatted text corresponding to the name

https://tools.ietf.org/html/rfc6350#section-6.2.1

khard/khard.py
```diff
@@ -474,18 +479,14 @@ def get_contacts(address_books, query, method="all", reverse=False,
...
-    :param method: the search method, one of "all", "name" or "uid"
-    :type method: str
-    :param reverse: reverse the order of the returned contacts
-    :type reverse: bool
-    :param group: group results by address book
-    :type group: bool
-    :param sort: the field to use for sorting, one of "first_name", "last_name"
-    :type sort: str
+    :param  address_books: the address books to search
+    :type address_books:
```
hydrargyrum: why did you remove the type here?

khard/carddav_object.py
```diff
@@ -1171,6 +1167,10 @@ def _process_user_input(self, input):
             contact_data.get("Prefix", ""), contact_data.get("First name", ""),
             contact_data.get("Additional", ""),
             contact_data.get("Last name", ""), contact_data.get("Suffix", ""))
+        if "Formatted name" in contact_data:
+            self.formatted_name = contact_data.get("Formatted name")
+        if not self.formatted_name:
+            self.formatted_name = ""
```
hydrargyrum: From reading the diff, I know this is supposed to trigger the `property.setter`, but if someone later on reads this code, they will probably miss the context and think this `if not x: x = ""` is useless, or worse, remove it. Maybe this could deserve a little code comment?

khard/khard.py
lucc: Editing mistake, fixed

khard/carddav_object.py
lucc: good idea

hydrargyrum:
@lucc  For some reasons, it seems the tests are failing.

d7415:
> @lucc For some reason, it seems the tests are failing.

That failure seems to have been fixed in 28366a9c3b553015c0145baaab6d8f23b140edc4. A rebase (or the eventual merge) should fix it.

Issue #225, opened 2019-10-08
TypeError: can only concatenate str (not "list") to str


chilledfrogs:
Really sorry if this issue is somehow already here, but try as I may to find it, I haven't, so posting a new issue...

Essentially, on pretty much any operation with `khard` requiring a list of more than 1 contact of some sort (so almost all operations), `khard` spits out the following traceback (the debug messages don't seem to show anything out of the ordinary, but if you guys need it, I will include them, just ask):
```
Traceback (most recent call last):
  File "/bin/khard", line 11, in <module>
    load_entry_point('khard==0.14.0', 'console_scripts', 'khard')()
  File "/usr/lib/python3.7/site-packages/khard/khard.py", line 1786, in main
    vcard_list = generate_contact_list(config, args)
  File "/usr/lib/python3.7/site-packages/khard/khard.py", line 703, in generate_contact_list
    args.strict_search if "strict_search" in args else False)
  File "/usr/lib/python3.7/site-packages/khard/khard.py", line 468, in get_contact_list_by_user_selection
    config.reverse(), config.group_by_addressbook(), config.sort)
  File "/usr/lib/python3.7/site-packages/khard/khard.py", line 494, in get_contacts
    contacts.extend(address_book.search(query, method=method))
  File "/usr/lib/python3.7/site-packages/khard/address_book.py", line 156, in search
    return list(search_function(query))
  File "/usr/lib/python3.7/site-packages/khard/address_book.py", line 84, in _search_all
    contact_details = contact.print_vcard()
  File "/usr/lib/python3.7/site-packages/khard/carddav_object.py", line 1664, in print_vcard
    if self.organisations:
  File "/usr/lib/python3.7/site-packages/khard/carddav_object.py", line 580, in organisations
    return self._get_multi_property("ORG")
  File "/usr/lib/python3.7/site-packages/khard/carddav_object.py", line 138, in _get_multi_property
    values.append(str(ablabel) + ": " + child.value)
TypeError: can only concatenate str (not "list") to str
```

Going indeed to line 137 of carddav_object.py and changing `child.value` to `str(child.value)` ***seems*** to fix the problem, and I don't notice any display issues, but I'm a bit curious as to why this happened in the first place, for the last couple releases...

Thanks in advance guys

d7415:
This looks like it's probably #221.

d7415:
[...]

PR #226, opened 2019-10-09
Use dict internally for fields with ABLABEL

### LINKED ISSUE #221: Handle ABLABELS on ORG, ADR and friends
This is a bug report extracted from #220.

Currently khard crashes if any attribute containing a list (like `ORG`) also has an ABLABEL attached.

d7415:
Also allow for ABLABELs in `_add_organisation`.

Fixes #221, #225.

Still needs handling for ADR (maybe others?) and unit tests. Feedback welcome, but I might not get back to this until next week, so here's a start.

~~In its current form, this also breaks sorting in` _get_multi_property`, which causes several tests to fail.~~

### DIFF
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -574,7 +574,10 @@ class VCardWrapper:
         # check if fn attribute is already present
         if not self.vcard.getChildValue("fn") and self.organisations:
             # if not, set fn to organisation name
-            org_value = helpers.list_to_string(self.organisations[0], ", ")
+            first_org = self.organisations[0]
+            if isinstance(first_org, dict):
+                first_org = list(first_org.values())[0]
+            org_value = helpers.list_to_string(first_org, ", ")
             self.formatted_name = org_value.replace("\n", " ").replace("\\",
                                                                        "")
             showas_obj = self.vcard.add('x-abshowas')
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -570,7 +570,8 @@ class VCardWrapper:
         :param str|list(str) organisation: the value to add
         :returns: None
         """
-        self._add_labelled_object("org", organisation, True, ObjectType.list_with_strings)
+        self._add_labelled_object("org", organisation, True,
+            ObjectType.list_with_strings)
         # check if fn attribute is already present
         if not self.vcard.getChildValue("fn") and self.organisations:
             # if not, set fn to organisation name
@@ -586,52 +587,47 @@ class VCardWrapper:
     @property
     def titles(self):
         """
-        :rtype: list(list(str))
+        :rtype: list(str or dict(str))
         """
         return self._get_multi_property("TITLE")
 
     def _add_title(self, title):
-        title_obj = self.vcard.add('title')
-        title_obj.value = convert_to_vcard("title", title, ObjectType.string)
+        self._add_labelled_object("title", title, True)
 
     @property
     def roles(self):
         """
-        :rtype: list(list(str))
+        :rtype: list(str or dict(str))
         """
         return self._get_multi_property("ROLE")
 
     def _add_role(self, role):
-        role_obj = self.vcard.add('role')
-        role_obj.value = convert_to_vcard("role", role, ObjectType.string)
+        self._add_labelled_object("role", role, True)
 
     @property
     def nicknames(self):
         """
-        :rtype: list(list(str))
+        :rtype: list(str or dict(str))
         """
         return self._get_multi_property("NICKNAME")
 
     def _add_nickname(self, nickname):
-        nickname_obj = self.vcard.add('nickname')
-        nickname_obj.value = convert_to_vcard("nickname", nickname,
-                                              ObjectType.string)
+        self._add_labelled_object("nickname", nickname, True)
 
     @property
     def notes(self):
         """
-        :rtype: list(list(str))
+        :rtype: list(str or dict(str))
         """
         return self._get_multi_property("NOTE")
 
     def _add_note(self, note):
-        note_obj = self.vcard.add('note')
-        note_obj.value = convert_to_vcard("note", note, ObjectType.string)
+        self._add_labelled_object("note", note, True)
 
     @property
     def webpages(self):
         """
-        :rtype: list(str)
+        :rtype: list(str or dict(str))
         """
         return self._get_multi_property("URL")
 
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -937,7 +937,7 @@ class YAMLEditable(VCardWrapper):
                     {ablabel: child.value} if ablabel else child.value)
         # sort private object lists
         for value in private_objects.values():
-            value.sort()
+            value.sort(key=multi_property_key)
         return private_objects
 
     def _add_private_object(self, key, value):
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -57,6 +57,13 @@ def convert_to_vcard(name, value, allowed_object_type):
                      " must be a string or a list with strings.")
 
 
+def multi_property_key(item):
+    if isinstance(item, dict) or isinstance(item, list):
+        return list(item)[0]
+    else:
+        return item
+
+
 class VCardWrapper:
     """Wrapper class around a vobject.vCard object.
 
@@ -136,8 +143,7 @@ class VCardWrapper:
                     values.append({ablabel: child.value})
                 else:
                     values.append(child.value)
-        #return sorted(values)
-        return values
+        return sorted(values, key=multi_property_key)
 
     def _delete_vcard_object(self, name):
         """Delete all fields with the given name from the underlying vCard.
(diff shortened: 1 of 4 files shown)


test/test_vcard_wrapper.py
```diff
@@ -463,7 +463,7 @@ def test_setting_and_getting_webpage_ablabel(self):
         wrapper._add_webpage({'github': 'https://github.com/scheibler/khard'})
         wrapper._add_webpage('http://example.com')
         self.assertListEqual(wrapper.webpages, [
-            'github: https://github.com/scheibler/khard',
+            {'github': 'https://github.com/scheibler/khard'},
             'http://example.com'])
 
     @unittest.expectedFailure
```
lucc: If you claim to fix #221 you have to fix this test somehow.

d7415: (and check for other candidates)

d7415:
Labelled `ADR` and `TEL` fields seem to work without further modification. I need to remind myself what needs to be tested for the unit test, then this should be ready for review.

test/test_vcard_wrapper.py
d7415: If there's a reason this needs to not be a dict (or somewhere I haven't expanded it) let me know and I'll have another look. This seemed to be the neatest way of keeping all the data between functions.

d7415: I see what you meant now. github's interface didn't help by hiding the test you were talking about!

lucc: This checks for a list because `VcardWrapper.organisation` is documented to be a list. I just improved the docstring of `_add_organisation`, check out 3447d658b784e29ca4e5f45dd88ee6c7335cca7e.

d7415:
Rebasing was neater than merging. This should be ready for review now. If I've missed anything let me know.

I *think* the reason `ADR`/`TEL` fields worked unchanged is because they already handled `dict`s for e.g. `CELL` and `WORK`.

test/test_vcard_wrapper.py
d7415: Ah *looks*

d7415: Is there any reason I can't change the relevant docstrings? to `list(list(str)) or list(dict(list(str)))`?

d7415: I might need to return to this tomorrow to get the types all straight in my head.

d7415: Also, I suppose I should probably use `add_labelled_object` for any field using `_get_multi_property`, and adjust their docstrings accordingly

d7415: Ok. `TITLE`, `ROLE`, `NICKNAME`, and `NOTE` now use `_add_labelled_object`, and I've changed their docstrings and some others to reflect the appropriate `:rtype`s.

If I'm going in the wrong direction, let me know.

khard/helpers.py
```diff
@@ -168,6 +168,11 @@ def convert_to_yaml(name, value, indentation, index_of_colon,
                             ' ' * (indentation+8), indent_multiline_string(
                                 inner, indentation+12,
                                 show_multi_line_character)))
+            elif isinstance(outer, dict):
+                # ABLABEL'd lists
+                for k in outer:
+                    strings += convert_to_yaml("- " + k, outer[k], indentation+4,
```
lucc: Can you please fix the indent on this one. It should be one of these
```py
ret = function_name(a, b,
		    c, d)
ret = function_name(
    a, b, c, d)
```
Check [pep8: indent](https://www.python.org/dev/peps/pep-0008/#indentation).

test/test_vcard_wrapper.py
```diff
@@ -464,11 +464,16 @@ def test_setting_and_getting_webpage_ablabel(self):
...
             'http://example.com'])
 
-    @unittest.expectedFailure
     def test_labels_on_structured_values(self):
         vcard = VCardWrapper(_from_file('test/fixture/vcards/labels.vcf'))
-        # TODO find a solution for issue #221
-        self.assertListEqual(vcard.organisations, ['Test Inc'])
+        self.assertListEqual(vcard.organisations, [{'Work': ['Test Inc']}])
+
+    def test_setting_and_getting_webpages(self):
```
lucc: this has nothing to do with webpages

test/test_vcard_wrapper.py
```diff
@@ -464,11 +464,16 @@ def test_setting_and_getting_webpage_ablabel(self):
...
+        vcard = create_test_vcard()
+        wrapper = VCardWrapper(vcard)
+        wrapper._delete_vcard_object("FN")
```
lucc: please make at least two tests for this: one to test setting and getting of some labeled attribute and one for reconstructing the fn from a labeled attribute.

khard/carddav_object.py
```diff
@@ -562,13 +570,15 @@ def _add_organisation(self, organisation):
         :param str|list(str) organisation: the value to add
         :returns: None
         """
-        org_obj = self.vcard.add('org')
-        org_obj.value = convert_to_vcard("organisation", organisation,
-                                         ObjectType.list_with_strings)
+        self._add_labelled_object("org", organisation, True,
```
lucc: fix this indent too please

khard/carddav_object.py
```diff
@@ -377,7 +384,8 @@ def _get_new_group(self, group_type=""):
             else:
                 return group_name
 
-    def _add_labelled_object(self, obj_type, user_input, name_groups=False):
+    def _add_labelled_object(self, obj_type, user_input, name_groups=False,
+                             allowed_object_type=ObjectType.string):
```
lucc: I know this method did not have a docstring before but if you can it would be nice to add a docstring.

khard/carddav_object.py
```diff
@@ -57,6 +57,13 @@ def convert_to_vcard(name, value, allowed_object_type):
                      " must be a string or a list with strings.")
 
 
+def multi_property_key(item):
```
lucc: a docstring for this one please

d7415: While writing the docstring here (proving its usefulness) I realised a couple of things. Chiefly, I'm currently returning the label to sort here, but I'm not sure whether that's sensible.

Returning the value would make more sense. Better still I think would be sorting by type so that labelled are either before or after unlabelled, then by label for labelled and content for unlabelled. I have an idea for that.

d7415:
> I had a first look through the changes.

Thanks! Sorry about the silly errors - I should have caught those.

>     1. `ABLABEL` is not specified in the rfcs so where is it actually defined? How do we mention our handling of `ABLABEL`s in the docs?

That's a good question. It seems to be more "accepted behaviour" that conforms to standards rather than an official standard.

>     2. Can a user add new labels when editing a contact in yaml? The template was not updated to mention this but I am not sure if we should mention it right away even if it is supported. This point needs some thinking.

I might have missed something, but where I've tested it, yes. I think I'd agree about documentation, at least until point (1) is resolved.

>     3. for some properties (e.g. email) we already support labels per the type parameter. How does this interact with the ABLABELs? (what if both are present? what if both are present and they are edited?)

Email, phone and postal addresses currently use the TYPE parameter for those types specified in `email_types_v3` etc and ABLABELs for anything else.

khard/carddav_object.py
lucc: returning the value sounds good

test/test_vcard_wrapper.py
d7415: I just want to check what the new tests are testing for.

The other tests in this block test basic ABLABELs with webpages and loading the labelled `ORG` from the test file. This one ensures that the FN reconstruction works from a labelled `ORG`.

There's also a test in `test_yaml.py` for loading labelled/unlabelled webpages from yaml input.

khard/carddav_object.py
d7415: I managed to make it sort unlabelled objects first (by value), then the labelled ones by label. Would you prefer sorting all of them by value? It's an easy change.

lucc: I don't really care and it is easy to change later.

d7415:
> Sadly editing is broken on this branch. Try to find a vcard with an ABLABEL: `grep -RC5 ABLABEL /you/addressbook` and then edit that card.

Weird. It's working for mine... Which field(s) are labelled in the broken ones?

lucc:
@d7415 I'm very sorry this was an error on my side. I just reinvestigated and it seems that our yaml generation is not round trip save. In my case I had a contact with a single line NOTE containing a colon. That results in 
```yaml
Note: foo: bar
```
in the yaml template which does not parse if read back in. (`Note: "foo: bar"` or 
```yaml
Note: | 
   foo: bar
```
would be fine).

Sadly that contact also had an ABLABEL and was the first one I tested.

d7415:
Ah, so `NOTE` (or anything with a value containing a ": " - iirc the space is required) need some special handling

lucc:
Yea but that is unrelated to this PR.

d7415:
> Yea but that is unrelated to this PR.

I have a quick patch that seems to work. It's a bit of a hack, but forces strings containing ": " to be treated as multiline (so they get the `|`). I'm not sure whether this has effects elsewhere with other fields, but do you want me to PR it?

PR #227, opened 2019-10-17
Force strings that look like yaml mapping values to be displayed as multiline strings

d7415:
As @lucc discovered in while testing #226, a string containing ": " will be output into a yaml as a normal string, but will then fail to parse because it looks like a yaml mapping value.

This patch works around that problem by forcing any string containing ": " to be output into yaml as a multiline string, where such sequences are ignored. I've also included a test for the round trip, which should hopefully work even if the implementation of the fix changes.

### DIFF
--- a/test/test_yaml.py
+++ b/test/test_yaml.py
@@ -5,12 +5,15 @@ import datetime
 from io import StringIO
 import unittest
 from unittest import mock
+import copy
 
 from ruamel.yaml import YAML
 
 from khard.carddav_object import CarddavObject, YAMLEditable
 
 from . import helpers
+
+import khard.helpers
 
 
 def create_test_card():
@@ -172,3 +175,31 @@ class UpdateVcardWithYamlUserInput(unittest.TestCase):
         data = to_yaml(data)
         card.update(data)
         self.assertEqual(card.formatted_name, fn)
+
+    def test_parse_field(self):
+        """Test round-trip of a field to/from YAML"""
+        card = create_test_card()
+        data = "First name: Nobody\n"
+        data += "\n".join(khard.helpers.convert_to_yaml("Note", "foobar", 0,
+                                                        5, True))
+        card.update(data)
+        self.assertListEqual(card.notes, ["foobar"])
+
+    def test_parse_field_with_colon(self):
+        """Test round-trip of a field containing ': ' to/from YAML"""
+        card = create_test_card()
+        data = "First name: Nobody\n"
+        data += "\n".join(khard.helpers.convert_to_yaml("Note", "foo: bar", 0,
+                                                        5, True))
+        card.update(data)
+        self.assertListEqual(card.notes, ["foo: bar"])
+
+    def test_vcard_round_trip(self):
+        """Test a VCARD can be converted to YAML and back unchanged"""
+        card = create_test_card()
+        card._add_organisation("ACME, Inc")
+        card._add_note("foo: bar")
+        card2 = copy.copy(card)
+        yaml = card.get_template()
+        card.update(yaml)
+        self.assertEqual(card.vcard.serialize(), card2.vcard.serialize())
--- a/khard/helpers.py
+++ b/khard/helpers.py
@@ -118,7 +118,7 @@ def convert_to_yaml(name, value, indentation, index_of_colon,
     :param str name: name of object (example: phone)
     :param value: object contents
     :type value: str, list(str), list(list(str)), list(dict)
-    :param str indentation: indent all by number of spaces
+    :param int indentation: indent all by number of spaces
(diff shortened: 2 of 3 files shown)


d7415:
I'm not sure why Travis isn't running. Maybe I'm just impatient. They passed locally though.

lucc:
No idea why github doesn't display it but the build passed: https://travis-ci.org/scheibler/khard/builds/600116816

test/test_yaml.py
```diff
@@ -189,3 +191,14 @@ def test_update_fn(self):
         data = to_yaml(data)
         card.update(data)
         self.assertEqual(card.formatted_name, fn)
+
+    def test_parse_field_with_colon(self):
+        card = create_test_card()
+        fn = 'me myself and i'
```
lucc: this test has nothing to do with `FN` so please remove that. Tests should be as short and simple as possible.

test/test_yaml.py
```diff
@@ -189,3 +191,14 @@ def test_update_fn(self):
...
+        data = {'Formatted name': fn}
+        data = to_yaml(data)
+        print(khard.helpers.convert_to_yaml("Note", "foo: bar", 0, 5, True))
```
lucc: Why is there a print here? I think `print()` should be avoided in libraries and tests.

test/test_yaml.py
```diff
@@ -189,3 +191,14 @@ def test_update_fn(self):
...
+        data += "\n".join(khard.helpers.convert_to_yaml("Note", "foo: bar", 0,
+                                                        5, True))
+        card.update(data)
```
lucc: What we actually should test for are two things (in two tests):
1. that convert_to_yaml handles strings with colons special
2. more important that a round trip of a vcard to yaml and back is idempotent
   1. in general
   2. specially for values with colons

test/test_yaml.py
d7415: Thought I'd removed all of those - sorry! `pytest` hides the outputs if the tests pass, which might have fooled me...

test/test_yaml.py
d7415: Will do. I think the FN needs to be set or `card.update()` will fail, but I can inline it.

test/test_yaml.py
d7415: So 1 test as above, but with an additional non-colon value, and another test for (1)?

d7415: I was deliberately avoiding making the test specific to this behaviour (re point 1), but I can do that.

d7415: I think that should all be covered now.

test/test_yaml.py
```diff
@@ -189,3 +191,21 @@ def test_update_fn(self):
...
+        """Test round-trip of a field to/from YAML"""
+        card = create_test_card()
+        data = "First name: Nobody\n"
+        data += "\n".join(khard.helpers.convert_to_yaml("Note", "foobar", 0,
+                                                        5, True))
+        card.update(data)
+        self.assertListEqual(card.notes, ["foobar"])
+
+    def test_parse_field_with_colon(self):
+        """Test round-trip of a field containing ': ' to/from YAML"""
```
lucc: You advertise these two tests as "round trip" but that's not what they do. A round trip test would look something like this:
```py
import copy
card = create_test_card()
# add extra data to the card if needed
card2 = copy.copy(card)
yml = card.convert_to_yaml()
card.process_user_inpu(yml)
assert card == card2
```
By round trip I mean that the data converted to yaml can safely be converted back without errors or changes. Like a user who did not actually edit the yaml file.

d7415: Ah, ok - you want to test the whole card is identical, but I was only testing one field (though it was being converted to yaml and back). I'll change that.

d7415: It's not super clean because the `card == card2` won't work (I guess we'd have to write handling of that into the class?) so I'm testing `get_template()`. Which, come to think of it, isn't great because that's YAML again...

d7415: Now using `serialize()` which makes more sense than testing the YAML.

lucc: `serialize` sounds good for this purpose but please do not import modules inside test methods (or functions or anywhere else but the top of the module if there is not a **very good** reason).  And please use the correct assert function from the unittest class.  My code was just an explanation what I had in mind.

lucc:
This looks ready apart from my last comment. Can you please rebase on newest the develop branch and edit your commit history into one meaningful story?

d7415:
> Can you please rebase on newest the develop branch and edit your commit history into one meaningful story?

I've squashed it into one, removing irrelevant commit text. If you'd like it broken down more I still have the original tree locally.

d7415:
Thanks! :)

PR #233, opened 2019-10-28
Make the build reproducible

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) effort we noticed that *khard* could not be built reproducibly.

This is because it includes the absolute build directory in the documentation via the `SPEC_FILE` attribute. This commit calculates this value dynamically instead.

(This was originally filed in Debian as [#943471](https://bugs.debian.org/943471).)

### DIFF
--- a/khard/config.py
+++ b/khard/config.py
@@ -92,7 +92,6 @@ def validate_private_objects(value):
 class Config:
 
     supported_vcard_versions = ("3.0", "4.0")
-    SPEC_FILE = os.path.join(os.path.dirname(__file__), 'data', 'config.spec')
 
     def __init__(self, config_file=None):
         self.config = None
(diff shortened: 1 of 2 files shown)


khard/config.py
```diff
@@ -92,7 +92,7 @@ def validate_private_objects(value):
 class Config:
 
     supported_vcard_versions = ("3.0", "4.0")
-    SPEC_FILE = os.path.join(os.path.dirname(__file__), 'data', 'config.spec')
+    SPEC_FILE = None
```
lucc: This is never read so it should be removed, including the test if it is None below.

khard/config.py
```diff
@@ -116,8 +116,12 @@ def _load_config_file(cls, config_file):
             config_file = os.getenv("KHARD_CONFIG", os.path.join(
                 xdg_config_home, "khard", "khard.conf"))
         try:
+            configspec = cls.SPEC_FILE
+            if configspec is None:
+                configspec = os.path.join(os.path.dirname(__file__),
+                                          'data', 'config.spec')
```
lucc: This does not belong in the try block.

lamby:
Thanks for the quick review; pushed updated commit..

lucc:
Thanks for you help @lamby !

PR #237, opened 2019-11-25
Add a category search with --category option.

vleonbonnet:


lucc:
Thank you @vleonbonnet but I think I will not merge this as it only covers a special case. I would prefer the solution discussed in #131. But can use this for reference! :)

(Sorry for the late reply)

PR #241, opened 2020-01-17
Works fine on macOS and the *BSDs

rjc:
The current wording may, falsely, imply that `khard` runs is Linux-only, which isn't the case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 khard
 =====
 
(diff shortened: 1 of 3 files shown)


rjc:
Commit updated to include all `Linux` -> `Unix` changes.

rjc:
Ping.

lucc:
Sorry I was busy with other stuff.

rjc:
No worries and thanks! :^)

Issue #204, opened 2019-01-25
Improve the documentation


lucc:
Currently we have a lot of documentation in the README. This issue is for planning and discussing further changes to improve the documentation.

The plan is to make the README smaller and move the documentation to the doc folder where we use sphinx. This was started in #201.

For the minimal README we want at least this information:
* short description of khard
* (pip) installation instructions
* a few usage examples.
* pointers to the documentation

I think the plan looks like this:
* [x] extend the sphinx documentation 
  * [x] configuration (#216)
  * [x] command line usage (#216)
  * [x] scripting (#229)
  * [x] integration with email clients and other programs (#216)
* [x] reduce the README (#216)
* [x] write man pages
  * [x] main man page for khard (#201)
  * [x] for the config file
* [x] publish the docs somewhere, ~e.g.~ [readthedocs](https://khard.readthedocs.io/en/latest/)

lucc:
We should also check if and how we can use the --help output to auto-generate a listing of all subcommands and their options. This should then be used in the man page and the command line usage section.

lucc:
https://sphinx-argparse.readthedocs.io/en/stable/index.html

lucc:
Maybe we could use typehints to document the parameter and return types: https://github.com/agronholm/sphinx-autodoc-typehints

PR #246, opened 2020-03-22
Command line interface: Implement --fields for list command

earboxer:
https://github.com/scheibler/khard/issues/179#issuecomment-576060333

Things you can do with this

* `python3 -m khard ls -F name,birthday` (like `khard bdays`, but it doesn't sort, doesn't exclude missing birthdays, and outputs in a different format)
* `python3 -m khard ls -pF emails.home.0,name` (shows first home email address for each user)
* `python3 -m  khard ls --fields phone_numbers.cell.0,phone_numbers` (shows first cell number, and all phone numbers in python format)
* `python3 -m khard ls -p -F Name,post_addresses.home.0.region,post_addresses.home.0.code` (name, state and zip code)

The list of properties can be seen with

`grep '@property' -A1 khard/carddav_object.py`

(Additional fields can easily be added by making them attributes in carddav_object.py)

For backwards compatibility with previous usage, the additional fields are defined:
* index (counts 1,2,3,...)
* name (formats name as desired)
* ~~address_book~~
* phone
* email
*  uid (prints minimized when not using --parsable)
* ~~uid_short~~

### DIFF
--- a/khard/khard.py
+++ b/khard/khard.py
@@ -375,7 +375,8 @@ def get_nested_field(vcard, field):
                 val = val[partial.upper()]
             else:
                 val = ''
-    return val
+    # Convert None and other falsy values to the empty string
+    return val or ''
 
 
 def get_special_field(vcard, field):
--- a/khard/khard.py
+++ b/khard/khard.py
@@ -298,6 +298,7 @@ def list_address_books(address_books):
 
 
 def pretty_name(name):
+    """Convert snake_cased column into their pretty alternative."""
     # Note: ugly_name should give the key for each of these.
     field = {
         'index': "Index",
@@ -313,6 +314,7 @@ def pretty_name(name):
 
 
 def ugly_name(name):
+    """Convert SOME field-names into their_snakecase format."""
     return name.replace('-', '').replace(' ', '_').lower()
 
 
@@ -400,6 +402,8 @@ def get_nested_field(vcard, field):
 
 
 def get_special_field(vcard, field):
+    """Returns certain fields with specific formatting options
+        (for support of some list command options)."""
     if field == 'name':
         if vcard.nicknames and config.show_nicknames:
             if config.display == "first_name":
@@ -1038,6 +1042,8 @@ def list_subcommand(vcard_list, parsable, fields):
     :type vcard_list: list of carddav_object.CarddavObject
     :param parsable: machine readable output: columns devided by tabulator (\t)
     :type parsable: bool
+    :param fields: list of strings for field evaluation
+    :type fields: list
     :returns: None
     :rtype: None
 
--- a/khard/khard.py
+++ b/khard/khard.py
@@ -297,6 +297,25 @@ def list_address_books(address_books):
     print(helpers.pretty_print(table))
 
 
+def pretty_name(name):
+    # Note: ugly_name should give the key for each of these.
+    field = {
+        'index': "Index",
+        'name': "Name",
+        'phone': "Phone",
+        'email': "E-Mail",
+        'address_book': "Address book",
+        'uid': "UID",
+    }
+    if name in field:
+        return field[name]
+    return name
+
+
+def ugly_name(name):
+    return name.replace('-', '').replace(' ', '_').lower()
+
+
 def list_contacts(vcard_list, fields=[]):
     selected_address_books = []
     for contact in vcard_list:
@@ -306,49 +325,62 @@ def list_contacts(vcard_list, fields=[]):
     # table header
     if len(selected_address_books) == 1:
         print("Address book: {}".format(selected_address_books[0]))
-        table_header = ["Index", "Name", "Phone", "E-Mail"]
+        table_header = ["index", "name", "phone", "email"]
     else:
         print("Address books: {}".format(', '.join(
             [str(book) for book in selected_address_books])))
-        table_header = ["Index", "Name", "Phone", "E-Mail", "Address book"]
+        table_header = ["index", "name", "phone", "email", "addressbook"]
     if config.show_uids:
-        table_header.append("UID")
+        table_header.append("uid")
     if fields:
-        field = {
-            'index': "Index",
-            'name': "Name",
-            'phone': "Phone",
-            'email': "E-Mail",
-            'addressbook': "Address book",
-            'uid': "UID",
-        }
-        table_header = [field[x] for x in fields if x in field]
+        table_header = [ugly_name(x) for x in fields]
 
     abook_collection = AddressBookCollection('short uids collection',
                                              selected_address_books)
 
-    table.append(table_header)
+    table.append([pretty_name(x) for x in table_header])
     # table body
     for index, vcard in enumerate(vcard_list):
         row = []
-        row.append(index + 1)
+        for field in table_header:
+            if field == 'index':
+                row.append(index + 1)
+            elif field in ['name', 'phone', 'email']:
+                row.append(get_special_field(vcard, field))
+            elif field == 'addressbook':
+                row.append(vcard.address_book.name)
+            elif field == 'uid':
+                if abook_collection.get_short_uid(vcard.uid):
+                    row.append(abook_collection.get_short_uid(vcard.uid))
+                else:
+                    row.append("")
+            else:
+                # TODO: Allow grabbing ordered elements like 'emails[0]'
+                row.append(vcard.__getattribute__(field))
+        table.append(row)
+    print(helpers.pretty_print(table))
+
+
+def get_special_field(vcard, field):
+    if field == 'name':
         if vcard.nicknames and config.show_nicknames:
             if config.display == "first_name":
-                row.append("{} (Nickname: {})".format(
-                    vcard.get_first_name_last_name(), vcard.nicknames[0]))
+                return"{} (Nickname: {})".format(
+                    vcard.get_first_name_last_name(), vcard.nicknames[0])
             elif config.display == "formatted_name":
-                row.append("{} (Nickname: {})".format(vcard.formatted_name,
-                                                      vcard.nicknames[0]))
+                return"{} (Nickname: {})".format(vcard.formatted_name,
+                                                 vcard.nicknames[0])
             else:
-                row.append("{} (Nickname: {})".format(
-                    vcard.get_last_name_first_name(), vcard.nicknames[0]))
+                return "{} (Nickname: {})".format(
+                    vcard.get_last_name_first_name(), vcard.nicknames[0])
         else:
             if config.display == "first_name":
-                row.append(vcard.get_first_name_last_name())
+                return vcard.get_first_name_last_name()
             elif config.display == "formatted_name":
-                row.append(vcard.formatted_name)
+                return vcard.formatted_name
             else:
-                row.append(vcard.get_last_name_first_name())
+                return vcard.get_last_name_first_name()
+    elif field == 'phone':
         if vcard.phone_numbers:
             phone_dict = vcard.phone_numbers
             # filter out preferred phone type if set in config file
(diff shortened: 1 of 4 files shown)


earboxer:
This doesn't refactor any of the existing commands (I'm not sure what our deprecation strategy is for them).

This doesn't break existing functionality of list command.

This doesn't add a way to access first_name. (I'm not sure how we want to design the property for names)

lucc:
Thank you that looks very cool! Can you add a proper doc comment to all your functions and check pep8 formatting, please (I think I saw some function arguments wothout spaces inbetween).

I will do a full review next.

earboxer:
What do you think about the pretty_name function? It exists for the purpose of backwards compatibility (and maybe for the header to look nice?) It also exists so that the fields can be case insensitive... but maybe we would prefer to show the snake_cased ugly name instead?

khard/khard.py
```diff
@@ -296,47 +296,118 @@ def list_address_books(address_books):
...
+                            val = val[partial]
+                        elif partial.isdigit() and isinstance(val, list) \
+                                and len(val) > int(partial):
+                            val = val[int(partial)]
+                        # TODO: Completely support case insensitive indexing
+                        elif isinstance(val, dict) and partial.upper() in val:
+                            val = val[partial.upper()]
+                        else:
+                            val = ''
+                row.append(val)
```
lucc: Maybe you could put this in a separate function `get_nested_field`.  That would make it easier to write unit tests for this as it seems like an important piece of new code. What do you think?

khard/khard.py
```diff
@@ -296,47 +296,115 @@ def list_address_books(address_books):
...
+        'index': "Index",
+        'name': "Name",
+        'phone': "Phone",
+        'email': "E-Mail",
+        'address_book': "Address book",
+        'uid': "UID",
+    }
+    if name in field:
+        return field[name]
+    return name
```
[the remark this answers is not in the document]
earboxer: Hmm.. It seems like I wasn't aware of that syntactic sugar. There are probably a few other places where I should have used that.

lucc:
The idea of `pretty_name` and `ugly_name` looks good at first but there are some edge cases that do not look nice. Try `khard ls -F n-a-----m-e`.  Maybe it is better to only use alphabetic characters in field names and normalize them with `str.lower`.

khard/khard.py
```diff
@@ -1030,6 +1034,8 @@ def list_subcommand(vcard_list, parsable, fields):
     :type vcard_list: list of carddav_object.CarddavObject
     :param parsable: machine readable output: columns devided by tabulator (\t)
     :type parsable: bool
+    :param fields: list of strings for field evaluation
+    :type fields: list
```
lucc: list(str)

khard/khard.py
earboxer: Done. In what file should the tests be written for this function?

earboxer:
> The idea of pretty_name and ugly_name looks good at first but there are some edge cases that do not look nice. Try khard ls -F n-a-----m-e. Maybe it is better to only use alphabetic characters in field names and normalize them with str.lower.

So then, should the "pretty" formats be TitleCase? (If the header reads "E-mail" as it does, then if someone sees that and wants to use `khard -F E-mail`, it wouldn't work). This would be a backwards incompatible change (though maybe those only matter for `--parsable`).

khard/khard.py
earboxer: Is this used to generate some documentation somewhere?

Otherwise, this type of repetitious documentation is not helping anyone,
would make the writers of https://www.python.org/dev/peps/pep-0257 sad.

Please update [CONTRIBUTING.rst](https://github.com/scheibler/khard/blob/develop/CONTRIBUTING.rst)
if this type of documentatino needs to be written.

(otherwise, feel free to add it yourself).

earboxer:
Thank you for all the prompt feedback!

earboxer:
> Maybe it is better to only use alphabetic characters in field names and normalize them with str.lower.

Another consideration is that it is pulling attribute names from carddav_object, of which, many have underscores

```
version
uid
birthday
anniversary
formatted_name
organisations
titles
roles
nicknames
notes
webpages
categories
phone_numbers
emails
post_addresses
```

khard/khard.py
lucc: These are used in the sphinx generated documentation.  Try `make -c doc html && $BROWSER doc/build/html/index.html` locally or go to https://khard.readthedocs.io/ .  Although the python modules (and these signatures) are not yet visible on readthedocs (only after the next release).

I tend to put a more "human" description in the `:param:` part and the explicit type signature in the `:type:` part. 

```suggestion
    :param fields: the object path to the field on the vcard that should be printed
    :type fields: list(str)
```

But I can fix these small things when I merge. I know that I am very picky about such things.

About duplication info: I am also thinking about real python type signatures in #204. But that is another big project and seems to bring its own complications.

khard/khard.py
lucc: I think you could start a new test class in `test/test_command_line_interface.py` which is named after the function and has tests for some important cases of its behaviour.

lucc:
Is there any special reason you invited my to your khard repository?

I was thinking about normalizing field names and spotting typos:

I suggest that we allow only a very minimal character set in field names for the `-F` option. Maybe only `[a-z]` or `[a-z_]`. But I suggest not to allow `-`. I see that the header line currently says "E-Mail" but Wikipedia for example spells it [Email](https://en.wikipedia.org/wiki/Email). And I think that the field names in the header line are only read by humans and not programs.  The `-p` format is modeled after mutt and they don't read the first line at all. They assume it is headers and skip it.

So if a user specifies invalid characters we can give an error message. But what about field names? Try `khard ls -F namee`. That will result in a bunch of empty lines which is ugly. I have two ideas:
1. If the fields for a matching contact result in only empty strings (so that the line would be empty) we could skip that contact in the output. So `khard ls -F email foo` would only output the email address of contacts matching `foo` *and* having an email to display. On the other hand `khard ls -F name,email foo` would output all names and emails of contacts matching `foo` that have *at least* a name *or* an email to show.
2. We could manage a list of valid fields (derived from the CarddavObject class maybe) and reject any other (top level) fields (nested fields might be harder).

Furthermore I was thinking about private object and other special cases:

I think we can postpone private objects and formatted post addresses and formatted anniversary and birthdays. But I came up with a plan that might be doable after this PR: Add new read only properties to the CarddavObject for these special cases (and other special cases like the different name gettes) so that they can be handled in a uniform way when formating output with fields. Later these can also be used for a custom query syntax like planned in #131.

lucc:
Is there a reason why you have a markdown list token in front of every commit? I would remove them.

lucc:
I like the idea about a fixed list of valid fields more the longer I think about it.  We could still ignore case but require one specific spelling (plural or singular, underscores) and given propper error messages.

earboxer:
> Is there a reason why you have a markdown list token in front of every commit? I would remove them.

When I'm committing to a repository that doesn't have a commit message standard, I just go with what I'm used to (what we use at my workplace, which I believe was inspired by [Code::Blocks](https://github.com/obfuscated/codeblocks_sf/commits/master)'s standard)

`-/* System Name: Describe what it does in imperative mood`

`-` means minor commit (documentation, one-line bugfix) `*` means major commit.

earboxer:
Right now `uid` appears differently if you use parsable than if you did not.

It is currently the only field that does so (and special logic was written so it does so).

Should we have a specific way of referring to the shortened uid, like `uid_short`, so that either option can be available in both parsable and non-parsable ways?

khard/khard.py
earboxer: Sounds like a good idea.

This merge request is getting pretty big though. I don't know when/if I'll get around to writing comprehensive tests for that function.

earboxer:
It should be noted that https://github.com/scheibler/khard/pull/246/commits/44c090d7f8020be001a2c7c688498a5582c6d65b broke the existing tests...

khard/khard.py
```diff
@@ -296,47 +296,103 @@ def list_address_books(address_books):
...
-        abook_collection = AddressBookCollection('short uids collection',
-                                                 selected_address_books)
+        table_header.append("uid_short")
+
+    if parsable:
+        # Legacy default header fields for parsable.
+        table_header = ["uid", "name", "address_book"]
+
+    if fields:
+        table_header = [x.lower() for x in fields]
```
lucc: I would suggest to use `x.capitalize().replace('_', ' ')` instead. This yields the old title in most cases and is generally a bit nicer for humans that the internal name.

lucc:
I would ask you to remove the `*` and `-` at the beginning of your commit messages. They might make sense if the people reading them know the policy but that is not (yet? :) the policy for this repository so they carry no information for most readers.

The category or section name in the beginning og the commit message is nice and can also be understood by new readers without extra documentation.

lucc:
Currently `None` is printed for the `birthday` field if no birthday is set on the contact. This is in contrast to some/all? other fields which print the empty string.  I would suggest to make it the empty string for all fields unless there are good arguments for `None`.

Aditionally we should think about two things related to empty fields (one was already mentioned above):
1. What happens with a contact that has all requested fields empty? I suggest to drop that contact (it would just produce an empty line).
2. Should we add some special field modifier to require a value in that field? Something like `khard ls --fields 'name,!birthday'` `'name,birthday!` or any other char we deem acceptable. This would drop any contact from the output that does not have that field set. This would allow us to implement `khard bdays` as `khard ls --fields 'name,birthday!'`. An argument against this is that it mixes output logic with search term logic and such a constraint might better be represented in a custom query syntax like proposed in #131.

earboxer:
> Currently None is printed for the birthday field if no birthday is set on the contact. This is in contrast to some/all? other fields which print the empty string. I would suggest to make it the empty string for all fields unless there are good arguments for None.

It's just doing python string conversion: the attribute is returning None. We should probably change the __str__() method on the value returned so it shows only the day, or the empty string.

> What happens with a contact that has all requested fields empty? I suggest to drop that contact (it would just produce an empty line).

`khard ls -F phone_numbers.cell | grep '\S'`

Show only lines that have non-space-characters. (We would definitely want our `None` and `{}` to print as empty string in that case)

> `khard ls --fields 'name,birthday!'`

`khard ls --parsable --fields name,birthday | grep -Pv '\tNone'`

-P for perl regex (allows the \t to mean tab), -v for inverse.

If we change it to output the empty string instead of 'None', this would be `grep -Pv '\t$' ($ meaning the end of the line).

I believe this project should only be as complicated as necessary... 
 `khard show` and `khard edit` should also reflect the attributes of the vcard object (so the fieldnames are consistent throughout). -- maybe the attributes of the vcard object should change to match `khard show`... (which would encourage delaying the merge/release of this feature/branch)

khard/khard.py
earboxer: 97a6568

earboxer:
> I do not understand the reason for a6c2e25 . Why do we need the explicit name?

Was probably a little too overzealous making breaking changes. -- the idea was that the presentation of a field should not change when you change the `-p` flag (what if you wanted the short uid in a tab-separated output?)

Thinking about it more, it's probably not worth it for having the header change.

 reverted in 98b395b

I think the only breaking chance you are authorizing is changing

* 'E-Mail' to 'Email'
* 'UID' to 'Uid'

earboxer:
For a web-based project, I have been thinking about Edit-List-View for different fields of a model. In **khard**, this is pretty similar (except that the command for view is called show, and we have multiple lists for pseudo-models email, postaddress, phone).

Really, `khard edit` and `khard show` should be identical, except for the comments provided by `edit`

khard/khard.py
```diff
@@ -320,13 +320,13 @@ def list_contacts(vcard_list, fields=[], parsable=False):
...
     if fields:
-        table_header = [x.lower() for x in fields]
+        table_header = [x.lower().replace(' ','_') for x in fields]
 
     abook_collection = AddressBookCollection('short uids collection',
                                              selected_address_books)
 
     if not parsable:
-        table.append(table_header)
+        table.append([x.title().replace('_', ' ') for x in table_header])
```
lucc: Does it make sense to replace "_" with " " after `title()`? Isn't it the same as `capitalize()` then? I suggest to eiter use `capitalize()` or replace before using `title()`.

earboxer: I didn't test in every python version, but I found

```
>>> "python_khard".title()
'Python_Khard'
>>> "python khard".title()
'Python Khard'
```

Does this behave differently in different python versions?

lucc: Mis-asumption on my side, sorry :(

earboxer:
Okay, I fixed the tests, and rewrote history a little.

lucc:
Thank you very much @earboxer :)

PR #247, opened 2020-03-27
--fields aftermath

lucc:
@earboxer  I made some more changes to the `--fields` related code.  Would you mind reviewing them?

### DIFF
--- a/khard/khard.py
+++ b/khard/khard.py
@@ -383,21 +383,15 @@ def get_special_field(vcard, field):
     """Returns certain fields with specific formatting options
         (for support of some list command options)."""
     if field == 'name':
-        if vcard.nicknames and config.show_nicknames:
-            if config.display == "first_name":
-                return "{} (Nickname: {})".format(
-                    vcard.get_first_name_last_name(), vcard.nicknames[0])
-            if config.display == "formatted_name":
-                return "{} (Nickname: {})".format(vcard.formatted_name,
-                                                  vcard.nicknames[0])
-            return "{} (Nickname: {})".format(
-                vcard.get_last_name_first_name(), vcard.nicknames[0])
+        if config.display == "first_name":
+            name = vcard.get_first_name_last_name()
+        elif config.display == "formatted_name":
+            name = vcard.formatted_name
         else:
-            if config.display == "first_name":
-                return vcard.get_first_name_last_name()
-            if config.display == "formatted_name":
-                return vcard.formatted_name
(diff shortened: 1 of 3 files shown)


earboxer:
Looks great! Though it took me a while before I understood what format_labeled_field was doing.

lucc:
@earboxer ok, thanks. 

I added bf83651 to make the function easier to understand. The function is actually just the repeated code from the phone and email formatting you wrote.

I also added some tests for the new function. But one case is puzzeling me. The commit 
edb0589 contains a test that demonstrates a strange behaviour I found in the code. The commit a53e686 contains what I suppose is the fix. Is there any special reason you used only the first character?

earboxer:
> phone and email formatting you wrote

I actually just moved it into its own function. (see b10aa5a2f16b7afbca909b253bc342862c37c6ef)

 af4e6d1c26d12688efe09f04fce3be1833e8005f gives some documentation, and 3ec76ea shows when it was originally written (by @scheibler > 4 years ago for v0.7.1).

The only guess I can hazard is that it's for performance. Probably would be best to drop it, since it's against the defined behavior:

> If there is still no match, sort the types alphabetically

lucc:
I think I it a left over from refactoring.  The commit 3ec76ea you mentioned sorts `dict.items()` instead of `dict.keys()` so the `[0]` picks the key for each tuple and not the first char of the key. See https://github.com/scheibler/khard/commit/3ec76ea02d6fb5395187d227e1c7889217b68bc8#diff-c8f0b781e30179c1af4b32ce761d72beR541

I found the same pattern on the current develop branch: https://github.com/scheibler/khard/blob/ed65d9e44dc05994916efaf0f7768472697db1ec/khard/khard.py#L728

Thanks for your help with reviewing and finding these old commits :)

earboxer:
> Thanks for your help with reviewing and finding these old commits :)

No problem. My workflow for doing this is

```
git blame khard/khard.py
# Find the commit where that line was written, examine it.
git show af4e6d1
# Determine that we need to search deeper
git blame af4e6d1~1 khard/khard.py
# Find the commit where that line was written, examine it.
# ... repeat
```

The workflow could be done a little faster using [`tig`](https://github.com/jonas/tig)


```
tig blame khard/khard.py
# Find the line was written, press enter to examine it, q to exit
# If we need to search deeper, press `,` (comma) or type `:parent` to view the blame for the previous commit
# Find the line...
```

PR #250, opened 2020-04-07
Fix #249

### LINKED ISSUE #249: crash on specific string
Hi,

Using khard with mutt (and without) gives me some strange behaviour :
```
$khard -c ~/.config/khard/khard.conf email ant
Traceback (most recent call last):
  File "/usr/bin/khard", line 11, in <module>
    load_entry_point('khard==0.16.1.dev3+ged5b451', 'console_scripts', 'khard')()
  File "/usr/lib64/python3.8/site-packages/khard/khard.py", line 1163, in main
    vcard_list = generate_contact_list(args)
  File "/usr/lib64/python3.8/site-packages/khard/khard.py", line 559, in generate_contact_list
    vcard_list = get_contact_list_by_user_selection(
  File "/usr/lib64/python3.8/site-packages/khard/khard.py", line 407, in get_contact_list_by_user_selection
    return get_contacts(address_books, search,
  File "/usr/lib64/python3.8/site-packages/khard/khard.py", line 448, in get_contacts
    return sorted(contacts, reverse=reverse, key=lambda x:
  File "/usr/lib64/python3.8/site-packages/khard/address_book.py", line 85, in _search_all
    and len(re.sub(r"\D", "", query)) >= 3:
  File "/usr/lib64/python3.8/re.py", line 208, in sub
    return _compile(pattern, flags).sub(repl, string, count)
TypeError: expected string or bytes-like object
```

This would not happen with longer string. For some specific, it works sometimes ("pap" gives results but "isa" fails, "pa" or cg fails and "is" gives results).
I noticed  the results are one per line, except sometimes (like with "is"), there's an empty line between each result.

Compiled from sources (git, release tarball would'nt work).

It seems that it's a strange bug shown with debug option : it fails because the searched string doesn't really exist in the vcard.
I search ever and it fails with a card saying namephilippeverXXX but having First name "Philippe" and laste name "EverXXXsomething" without the real string in the vcard.
Same fail with Jan Toxic while searching for "anto".

lucc:
Closes #249


Tonus1:
Fine for me : tested a few crash strings with only the fix and everything is ok, thank you !

PR #255, opened 2020-04-18
Fix #251 part 2

### LINKED ISSUE #251: [Regression] Nickname occurs in the output of `khard email --parsable`
As per the [recommendations](https://khard.readthedocs.io/en/latest/scripting.html#mutt), I have the following in `~/.mutt/muttrc`:
```
set query_command="khard email --parsable %s"
bind editor <Tab> complete-query
```
When I attempt to invoke khard in neomutt, the nickname also appears in the recipient field. For example given the following vcard:
```
Formatted name : Joe Citizen
First name : Joe
Last name  : Citizen
Nickname : horrible_human_being
```
Neomutt addresses the email to 
```
"Joe Citizen (Nickname: horrible_human_being)" <jcitizen@foo.com>
```

This occurs with the current khard 0.16.1 (and 0.16.0), but this behaviour was not present in 0.15.1. I'm using python 3.8.2.

FWIW I notice other improvements in 0.16.0. Firstly, in both 0.15.1 and 0.16.0/0.16.1, I can search by nickname perfectly fine in general. However, the searching behaviour is slightly better in 0.16.0/0.16.1 if there are conflicting matches. Let's say I have two vcards:

1. `Michael Smith`, with nickname `Mike`
2. `Mike Jones`

With 0.16.0 or 0.16.1, searching for `Mike` will bring up both options in the selection dialogue.

However, with 0.15.1, it immediately auto-completes to `Mike Jones`, with no option to select `Michael Smith`.

lucc:
@protist does it work for you?

Fixes #251.

### DIFF
--- a/khard/khard.py
+++ b/khard/khard.py
@@ -716,28 +716,19 @@ def phone_subcommand(search_terms: Query, vcard_list: List[CarddavObject],
     formatter = Formatter(config.display, config.preferred_email_address_type,
                           config.preferred_phone_number_type,
                           config.show_nicknames, parsable)
-    all_phone_numbers_list = []
-    matching_phone_number_list = []
+    numbers = []
     for vcard in vcard_list:
         for type, number_list in sorted(vcard.phone_numbers.items(),
                                         key=lambda k: k[0].lower()):
             for number in sorted(number_list):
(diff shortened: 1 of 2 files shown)


protist:
Yes, this works perfectly! Thank you!

Issue #257, opened 2020-06-10
Add-email from fields other than From


jeffa5:
The `add-email` command is really useful for piping emails to and adding the contact in the From field. I think it would also be useful to be able to extract contacts from other Header fields, e.g. Cc and To.

Since these can tend to be lists it might need an interactive element to be able to skip some contacts in those lists if the user doesn't want to add them, potentially displaying a numbered list of them and letting the user select numbers they want to add?

I guess this could be added behind some new flags to the `add-email` command, maybe `--cc` and `--to`, or a more general `-H <Header>`?

lucc:
We could add a `--all` option to iterate over all email addresses that can be found in the email message and for each ask the user if they want to import that address and look for a contact to add it to.

We could refactor the `add_email_subcommand` function into one function that adds emails to contacts and one that parses the email and asks the user for confirmation and then hands the email to the first function.  This first function would copy most of the code from the current `add_email_subcommand`.

PRs welcome :)

jeffa5:
Sounds great, I'll taking a look at implementing this :)

PR #258, opened 2020-06-11
Feature: add-email custom headers

jeffa5:
Fixes: https://github.com/scheibler/khard/issues/257

This adds new arguments to the `add-email` sub command, `-H`, `--header` and `--all`, which allow extracting the email addresses from the given header, or all of the headers if requested.

This also adds a loop over the found email addresses so that the user is given the opportunity to add each to a contact.

A later extension of this might allow `-H` and `--header` to accept a list of headers to use, but this isn't implemented here.

### DIFF
--- a/khard/cli.py
+++ b/khard/cli.py
@@ -40,6 +40,15 @@ def field_argument(orignal: str) -> List[str]:
     return ret
 
 
+def comma_separated_argument(original: str) -> List[str]:
+    """Return the original string split by commas
+
+    :param original: the value from the command line
+    :returns: the original value split at "," and lower cased
+    """
+    return [f.lower() for f in original.split(",")]
+
+
 def create_parsers() -> Tuple[argparse.ArgumentParser,
                               argparse.ArgumentParser]:
     """Create two argument parsers.
@@ -278,6 +287,14 @@ def create_parsers() -> Tuple[argparse.ArgumentParser,
     add_email_parser.add_argument(
         "--vcard-version", choices=("3.0", "4.0"), dest='preferred_version',
         help="Select preferred vcard version for new contact")
+    add_email_parser.add_argument(
+        "-H",
+        "--headers",
+        dest='fields',
+        default=["from"],
+        type=comma_separated_argument,
+        help="Extract contacts from the given comma separated header fields. \
+                `all` searches all headers.")
     subparsers.add_parser(
         "merge",
         aliases=Actions.get_aliases("merge"),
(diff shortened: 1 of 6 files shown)


khard/khard.py
```diff
@@ -543,37 +544,27 @@ def new_subcommand(selected_address_books: AddressBookCollection,
     else:
         create_new_contact(selected_address_book)
 
+def add_email_to_contact(name: str, email_address: str, abooks: AddressBookCollection) -> None:
+    """Add a new email address to the given contact, creating the contact if necessary.
```
lucc, line 548: PEP-8 line length please.

test/test_khard.py
```diff
@@ -44,3 +46,39 @@ def test_no_search_terms_result_in_any_queries(self):
                              source_search_terms=query.AnyQuery(),
                              target_contact=query.AnyQuery())
         self.assertEqual(expected, prepared["foo"])
+
```
lucc, line 49: PEP-8 line length please.

khard/cli.py
```diff
@@ -278,6 +278,12 @@ def create_parsers() -> Tuple[argparse.ArgumentParser,
     add_email_parser.add_argument(
         "--vcard-version", choices=("3.0", "4.0"), dest='preferred_version',
         help="Select preferred vcard version for new contact")
+    add_email_parser.add_argument(
+        "-H", "--header", dest='field', default="From", help="Extract contacts from the given header field"
+        )
+    add_email_parser.add_argument(
+        "--all", dest="field", action="store_const", const="all", help="Extract contacts from all headers"
+        )
```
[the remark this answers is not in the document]
jeffa5: Sure, I'll run a formatter over it.

Perhaps we could just have the `-H` option for now with the comment in the docs. I'd find it easier adding a new argument later rather than removing one.

Yes, I'm just having a play around with the cli parsing now. I've just checked the field fetching code and it converts both to lower before comparison so we should be safe here.

test/test_khard.py
jeffa5: Yep, is there a standard auto-formatter you'd recommend for this project?

jeffa5:
I've removed the `--all` and implemented the multiple fields for `-H`. If one of the fields is `all` then all headers are searched. This also lower cases all fields given to aid in error messages.

test/test_khard.py
lucc: No, I have a language server for python in my editor and have a marker for overlong lines.

PR #260, opened 2020-06-15
Offer the first and last names to the user to use

jeffa5:
When adding contacts through the add-email functionality it can give the
name of the contact and then ask to create it. It would be easier to add
the contacts if we at least guess the name parts.


jeffa5:
Thanks, I've updated the commit message :)

PR #261, opened 2020-06-16
Add failing case for search by email

jeffa5:
This mimics what happens in `add_email_to_contact`. The search there only
looks for names, not emails even though we have the email available.

### DIFF
--- a/test/test_khard.py
+++ b/test/test_khard.py
@@ -6,7 +6,7 @@ from unittest import mock
 
 from email.headerregistry import Address
 
-from khard import khard, query
+from khard import khard, query, address_book
 from khard.khard import find_email_addresses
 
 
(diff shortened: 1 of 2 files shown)


jeffa5:
I'd propose adding a new search method for `email` so that the `add-email` contact search can be more intelligent if it doesn't find one by the name.

test/test_khard.py
```diff
@@ -161,3 +161,14 @@ def test_find_email_addresses_finds_all_emails_with_other_headers_too(
                 username="mjones",
                 domain="machine.example")]
         self.assertEqual(expected, addrs)
+
+class TestSearchContacts(unittest.TestCase):
```
lucc, line 165: Normally I name the test classes after the method they test and the individual test methods name describe *what* they test more precicely. Can you fix the names here please.

test/test_khard.py
```diff
@@ -161,3 +161,14 @@ def test_find_email_addresses_finds_all_emails_with_other_headers_too(
...
+
+    def test_get_contact_list_by_user_selection(self):
```
lucc, line 167: When we want to merge this we need to mark it as an expected failure. Use the `unittest.expectedFailure` decorator.

jeffa5:
I'd originally just meant this as an example of a vcard which `khard` prompts me to add again due to the different name but existing email. I'm not going to pursue adding it as a test case at the moment.

PR #262, opened 2020-06-17
Add searching by email for add-email

### LINKED ISSUE #259: add-email: try looking up contact by email address
When adding an email from raw email headers the lookup is currently by `name` or `all` if strict isn't set. This doesn't seem to search the email field though so with a contact of the form:

`A <a@a.com>`
`A last <a.last@b.com>`

Where both contacts are the same person, even with those emails in the contact, it still says would I like to create a new contact for `A last`, when my stored contact name is just `A`.

jeffa5:
When a contact isn't found by the name we try and search for it by email
address instead. This covers the case when a contact can have multiple
addresses associated with it but a different name is used.

Fixes #259

### DIFF
--- a/khard/address_book.py
+++ b/khard/address_book.py
@@ -97,6 +97,23 @@ class AddressBook(metaclass=abc.ABCMeta):
             if query.match(contact.formatted_name):
                 yield contact
 
+    def _search_emails(self, query: Query) -> Generator[
+            "carddav_object.CarddavObject", None, None]:
+        """Search in the email field for contacts matching query.
+
+        :param query: the query to search for
+        :yields: all found contacts
+        """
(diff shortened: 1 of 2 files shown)


lucc:
Sorry for answering so late. I was thinking about this some time ago but then got distracted from khard.

My first problem with this PR is to understand why I want to add an email address to an account that matches that exact email address.  That sounds to me as if the email address is already there, so why add it again? Or do you intend something else?

jeffa5:
Its been a while but I think this was for the case where the same contact may have different names and multiple emails. So you may search by name and not find the person even though the email is already associated with a different name (for the same person).

I don't have any time to work more on this so I'm happy to close it.

Issue #263, opened 2020-06-20
Test failure with Python 3.8


dotlambda:
The testsuite currently results in
```
======================================================================
FAIL: test_global_help (test.test_command_line_interface.HelpOption)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/nix/store/n0b076p351ma864q38if4yglsg99hw2s-python3-3.8.3/lib/python3.8/unittest/mock.py", line 1325, in patched
    return func(*newargs, **newkeywargs)
  File "/build/khard-0.16.1/test/test_command_line_interface.py", line 48, in test_global_help
    self._test(['-h'], 'usage: TESTSUITE [-h]')
  File "/build/khard-0.16.1/test/test_command_line_interface.py", line 45, in _test
    self.assertTrue(text.startswith(expect))
AssertionError: False is not true
```
Thus I patched the testsuite to use `assertRegex`:
```py
@mock.patch('sys.argv', ['TESTSUITE'])
class HelpOption(unittest.TestCase):

    def _test(self, args, expect):
        """Test the command line args and compare the prefix of the output."""
        with self.assertRaises(SystemExit):
            with mock_stream() as stdout:
                cli.parse_args(args)
        text = stdout.getvalue()
        self.assertRegex(text, expect)

    def test_global_help(self):
        self._test(['-h'], r'^usage: TESTSUITE \[-h\]')
```
and get
```
======================================================================
FAIL: test_global_help (test.test_command_line_interface.HelpOption)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/nix/store/n0b076p351ma864q38if4yglsg99hw2s-python3-3.8.3/lib/python3.8/unittest/mock.py", line 1325, in patched
    return func(*newargs, **newkeywargs)
  File "/build/khard-0.16.1/test/test_command_line_interface.py", line 48, in test_global_help
    self._test(['-h'], r'^usage: TESTSUITE \[-h\]')
  File "/build/khard-0.16.1/test/test_command_line_interface.py", line 45, in _test
    self.assertRegex(text, expect)
[...]

PR #264, opened 2020-06-20
use regex to test help option

dotlambda:


lucc:
I will put this on hold until we are clear about #263 .

test/test_command_line_interface.py
```diff
@@ -42,17 +42,17 @@ def _test(self, args, expect):
             with mock_stream() as stdout:
                 cli.parse_args(args)
         text = stdout.getvalue()
-        self.assertTrue(text.startswith(expect))
+        self.assertRegex(text, expect)
 
     def test_global_help(self):
-        self._test(['-h'], 'usage: TESTSUITE [-h]')
+        self._test(['-h'], r'^usage: TESTSUITE\s+\[-h\]')
```
dotlambda, line 48: Would this be acceptable?
```suggestion
        self._test(['-h'], r'^usage: TESTSUITE \[-h\]')
```
It would provide a way of seeing what the difference between the two strings is.

lucc: Yes, that would be acceptable and the argument to see the diff when the test fails is good. Hopefully your sandbox in #263 is *not* fixed by this becaue I think it is another issue and I would not want to fix it with some random edit.

I think with this we can merge it.

Can we unlink #263 somehow? I don't know if you want to keep it open.

dotlambda: It does not fix #263 but it makes it easier to add a patch that does to nixpkgs.

lucc:
I took the liberty to expand your commit message and merge that: 60d4d64237426fd

PR #269, opened 2020-07-27
Expand env vars for addressbooks path config

jaki:
Expand environment variables for the `addressbooks.<name>.path` config
values.  With this, you can set an addressbook path to
`$XDG_DATA_HOME/contacts/foo`, for example.

This may be backwards incompatible for users that use literal dollar
characters (`$`) in their paths.  If the environment variable expansion
is possible, it will be unintentionally expanded.  If it is not
possible, however, it will not be expanded.  So the compatibility issue
is only when the path is expandable (e.g.
`~/classes/english/$LANG/contacts`).  With this change, I do not know of
a way to express these kinds of paths (escaping doesn't work).  Yet, I
doubt anyone would come across this issue, let alone have a literal
dollar sign in their path.

### DIFF
--- a/test/test_address_book.py
+++ b/test/test_address_book.py
@@ -122,21 +122,21 @@ class VcardAddressBookLoad(unittest.TestCase):
             'test could not be parsed.')
 
     @mock.patch.dict("os.environ", clear=True)
-    def test_no_expand_unset_env_var(self):
+    def test_do_not_expand_env_var_that_is_unset(self):
         # Unset env vars shouldn't expand.
         with self.assertRaises(FileNotFoundError):
             address_book.VdirAddressBook(
                 "test", "test/fixture/test.abook${}".format("KHARD_FOO"))
 
     @mock.patch.dict("os.environ", KHARD_FOO="")
-    def test_expand_set_env_var_empty(self):
+    def test_expand_env_var_that_is_empty(self):
         # Env vars set to empty string should expand to empty string.
         abook = address_book.VdirAddressBook(
             "test", "test/fixture/test.abook${}".format("KHARD_FOO"))
         self.assertEqual(abook.path, "test/fixture/test.abook")
 
     @mock.patch.dict("os.environ", KHARD_FOO="test/fixture")
-    def test_expand_set_env_var_nonempty(self):
+    def test_expand_env_var_that_is_nonempty(self):
         # Env vars set to nonempty string should expand appropriately.
         abook = address_book.VdirAddressBook(
             "test", "${}/test.abook".format("KHARD_FOO"))
--- a/test/test_address_book.py
+++ b/test/test_address_book.py
@@ -123,15 +123,10 @@ class VcardAddressBookLoad(unittest.TestCase):
 
     @mock.patch.dict("os.environ", clear=True)
     def test_no_expand_unset_env_var(self):
-        var_name = "KHARD_FOO"
         # Unset env vars shouldn't expand.
-        try:
+        with self.assertRaises(FileNotFoundError):
             address_book.VdirAddressBook(
-                "test", "test/fixture/test.abook${}".format(var_name))
-            self.assertTrue(False)
-        except FileNotFoundError as err:
-            self.assertIn("test/fixture/test.abook${}".format(var_name),
-                          err.args[0])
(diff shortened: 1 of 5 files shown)


jaki:
Let me add tests...

jaki:
Added tests; ready for review.

test/test_address_book.py
```diff
@@ -120,6 +121,30 @@ def test_unparsable_files_can_be_skipped(self):
             'WARNING:khard.address_book:1 of 1 vCard files of address book '
             'test could not be parsed.')
 
+    def test_env_var_paths(self):
+        var_name = "KHARD_FOO"
+
+        # Unset env vars shouldn't expand.
+        if var_name in os.environ:
+            os.environ.pop(var_name)
```
lucc, line 129: I think it is better to use `unittest.mock.patch` to mock environment variables during tests. There are some examples of this in the tests.

test/test_address_book.py
```diff
@@ -120,6 +121,30 @@ def test_unparsable_files_can_be_skipped(self):
...
+        try:
+            address_book.VdirAddressBook(
+                "test", "test/fixture/test.abook${}".format(var_name))
+            self.assertTrue(False)
+        except FileNotFoundError as err:
```
lucc, line 134: You can use `self.assertRaises` for this, there are examples in the code.

test/test_address_book.py
```diff
@@ -120,6 +121,30 @@ def test_unparsable_files_can_be_skipped(self):
...
+            self.assertIn("test/fixture/test.abook${}".format(var_name),
+                          err.args[0])
+
+        # Env vars set to empty string should expand to empty string.
+        os.environ[var_name] = ""
+        address_book.VdirAddressBook(
+            "test", "test/fixture/test.abook${}".format(var_name))
```
lucc, line 141: This should be an extra test method and have an assert.

test/test_address_book.py
```diff
@@ -120,6 +121,30 @@ def test_unparsable_files_can_be_skipped(self):
...
+
+        # Env vars set to nonempty string should expand appropriately.
+        os.environ[var_name] = "test/fixture"
+        address_book.VdirAddressBook(
```
lucc, line 145: Again, extra method and asset missing.

jaki: I'm not sure what you mean by "asse[r]t missing".  I instead added an assert to make sure that the abook got created properly.

jaki:
I missed updating the documentation...

test/test_address_book.py
lucc: I meant that you did execute some code from khard but did not assert anything
afterwards.

In my opinion the generic description of how a test case looks like is
```python
def test_whatever(self):
    # prepare some objects that we might need
    x, y, z = 1, 2, 3
    # set up some mocks if we need them
    with unittest.mock.patch(...):
	# execute the code we want to test
        result = the_function_to_test(x, y, z)
    # check that the correct thing happend
    self.assertEqual(result, "what I expected")
```

And a unit test should only check *one* thing. So it is better to make small
test functions. But you did already split the test function so I think you
understood.

lucc:
Thank you!

jaki:
I tested out how order of expansion works on a shell, and the behavior seems to be different.

bash:

```sh
FOO=root
ls ~$FOO
```

```
ls: cannot access '~root': No such file or directory
```

zsh:

```sh
FOO=root
ls ~$FOO
```

```
ls: cannot open directory '/root': Permission denied
```

bash/zsh:

```sh
BAR='~root'
ls $BAR
```

```
ls: cannot access '~root': No such file or directory
```

python:

```py
import os
os.environ["FOO"] = "root"
os.environ["BAR"] = "~root"
print(os.path.expanduser(os.path.expandvars("~$FOO")),
      os.path.expanduser(os.path.expandvars("$BAR")))
```

```
'/root /root'
```

I don't currently know of a way to fix this.  It's minor and unlikely to get encountered, so I think it's fine to just document it here.

lucc:
I consider these special to these shells and not relevant to khard.  The shell might be the inspiration for this PR but you did not even mention that this feature works "like the shell" in the docs or the commit message :)

PR #281, opened 2020-12-17
s/weather/whether/ in khard.conf.rst

mstmob:


d7415:
Good spot, but bad correction. It should be "whether".

lucc:
Thank you!

PR #297, opened 2021-07-23
Reduce empty address lines in "details" output

sciurius:
When formatting the postal address, empty fields cause empty lines and sometimes weird appearance.
For example:
````
Address
    HOME: 
        Main street 20

        1234 AB Blaricum
        , Netherlands
````
This PR removes empty fields from the address structure so they do not cause misbehaviour.

After the patch:
````
Address
    HOME: 
        Main street 20
        1234 AB Blaricum
        Netherlands
````


scheibler:
Thank you

Issue #276, opened 2020-10-14
Normalise (quote) strings when editing through the Yaml interface?


lockywolf:
I am opening a vcard that has passed the following route: khard(new)-vdirsyncer->Google-vdirsyncer->khard(edit).

```
BEGIN:VCARD
VERSION:3.0
N:KhardLastName;KhardTestName;;;
FN:KhardTestName KhardLastName
TITLE:PhD
ORG:KharderProject;Unit2
REV:2020-10-14T01:40:56Z
UID:7bb2d3d99283d675
TEL;TYPE=CELL:+0 000 0000000
TEL;TYPE=HOME:+0 000 1000000
NOTE:Khard note line 1. Khard note line 2.
item1.IMPP;X-SERVICE-TYPE=Skype:skype:from-phone-skype
X-DOMAIN-NAME:@khard-rss
X-FACEBOOK:@khard-facebook
X-JABBER:jabber@khard.er
X-MASTODON:@khard-mastodon
X-SKYPE:skype-kharder
X-TELEGRAM:@khard-telegram
X-TWITTER:@khard-twitter
item1.X-ABLabel:Other
END:VCARD
```

Too bad I didn't save the original .vcf created by khard, but the following is the issue:

I cannot edit this vcard with Yaml, because of the "@uid" values, as apparently Yaml tokens cannot begin with a "@" symbol. May I suggest enquoting values fed from vcf to yaml, unless already enquoted?
The other way would be ambiguous, but I think that removing the quotes is a better option.

lucc:
Currently conversion to yaml is implemented by hand in the methods `khard.carddav_object.YAMLEditable.to_yaml` and `khard.helpers.convert_to_yaml`.  We could fix it there but the clean solution would be to use a library for generating yaml. (We are using `ruamel.yaml` to *parse* yaml.)   If somebody wants to propose a library or even a PR that would be interesting.

I think the requirements for generating yaml with a library are that we can put comments into the output.  We currently have that for documentation. There are proposals (#146) for less verbose yaml output but that should just be an option for advanced users. For newcomers it should be as well documented as it is now.  The special indent rules and special case formatting might not be so important to keep.

skowalak:
I just tested reading and then writing the template with `ruamel.yaml` and the following script. It preserved the comments. Unfortunately, I am not very familiar with `khard`s codebase yet, but I could try creating a PR to use `ruamel.yaml` and quote yaml values.

``` python
import sys
from ruamel import yaml

[...]

Issue #146, opened 2017-11-28
Make YAML template configurable


mxmehl:
When adding a new contact, the full template (from [helpers.py#L263](https://github.com/scheibler/khard/blob/master/khard/helpers.py#L263)) is a bit too crowded for experienced users – I for myself don't need many of the fields.

Can we make this template configurable, either by a command line parameter or in khard.conf to use another plain yaml file? Right now the only way to edit this is by changing the helpers.py file before compilation which is not user-friendly.

whnr:
For the time being, try this as a one liner in the terminal of your choice:

    tmp=$(mktemp tmp-vcard-XXXXXX.yaml); cp short.yaml $tmp; vim $tmp; khard add -i $tmp; rm $tmp

I made an alias out of it.

My `short.yaml` looks like this:
```yaml
First name : 
Last name  : 

Birthday : 

Organisation : 
Title : 

Phone :
    cell : 
    work : 
    home : 

Email :
    home : 
    work : 

Address :
    home :
        Extended : 
        Street   : 
        Code     : 
        City     : 
        Region   : 
        Country  : 

# notes
# for multi-line notes use:
#   Note : |
#       line one
#       line two
Note : 
````

lucc:
I only gave my +1 above but forgot to say a bit more: I think it is a good idea. I would go with a config option (and maybe also a command line option additionally). The default value could be `None` which selects the old builtin template. Later we could also ship more templates, but that is of very low priority right now.

*The only think missing is that someone sends a pull request ;)*

lucc:
The corresponding code that we have right now is rather ugly. One part is in `khard.helpers.get_new_contact_template()` and the even uglier bit in `khard.carddav_object.CarddavObject.get_template()`.  Maybe we can find way to use `str.format()` to interpolate a given vcard into a yaml template. That would enable us to reuse the technique for custom output format strings on the command line.

Maybe simple python format strings might be to limited to cope with all the special things we currently do in our yaml template. In this case we could check out a templating engine like jinja2, but I would be happier if we can do without that extra complexity (for the end user).

lucc:
I have just moved the template string from helpers.py to a separate data file. We can now think about dropping more templates there and making them selectable and adding some other path (`~/.config/khard/templates`?) to the search path.

But there is one important thing we have to check with any new template and I have not yet investigated the corresponding code: **Does loading of any custom / non default yaml template change the vcard in unintended ways?**

To answer this question we need to define a behaviour for:
[...]

PR #302, opened 2022-02-27
Use ruamel.yaml for YAMLEditable

skowalak:
replace the YAMLEditable.to_yaml() method with ruamel serialization.
migrate some yaml formatting to the helpers module.

### DIFF
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -18,6 +18,7 @@ import time
 from typing import Callable, Dict, List, Optional, Tuple, Union, Sequence
 
 from atomicwrites import atomic_write
+from ruamel import yaml
 from ruamel.yaml import YAML
 import vobject
 
@@ -1257,9 +1258,9 @@ class YAMLEditable(VCardWrapper):
             "Categories": self.categories,
             "Note": self.notes,
             "Webpage": self.webpages,
-            "Anniversary": 
+            "Anniversary":
                 helpers.yaml_anniversary(self.anniversary, self.version),
-            "Birthday": 
+            "Birthday":
                 helpers.yaml_anniversary(self.birthday, self.version),
             "Address": helpers.yaml_addresses(
                 self.post_addresses, ["Box", "Extended", "Street", "Code",
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -1268,7 +1268,6 @@ class YAMLEditable(VCardWrapper):
         }
         template = helpers.get_new_contact_template()
         yaml = YAML()
-        yaml.top_level_colon_align = True
         yaml.indent(mapping=4, sequence=4, offset=2)
         template_obj = yaml.load(template)
         for key in template_obj:
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -1270,9 +1270,9 @@ class YAMLEditable(VCardWrapper):
         yaml.top_level_colon_align = True
         yaml.indent(mapping=4, sequence=4, offset=2)
         template_obj = yaml.load(template)
-        for k in template_obj:
-            v = translation_table.get(k, None)
-            template_obj[k] = helpers.yaml_clean(v)
+        for key in template_obj:
+            value = translation_table.get(key, None)
+            template_obj[key] = helpers.yaml_clean(value)
 
         if self.supported_private_objects:
             template_obj["Private"] = helpers.yaml_clean(
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -8,6 +8,7 @@ which can be found here:
 
 import copy
 import datetime
+import io
 import locale
 import logging
 import os
@@ -17,7 +18,7 @@ import time
 from typing import Callable, Dict, List, Optional, Tuple, Union, Sequence
 
 from atomicwrites import atomic_write
-from ruamel import yaml
+from ruamel.yaml import YAML
 import vobject
 
 from . import address_book  # pylint: disable=unused-import # for type checking
@@ -964,7 +965,7 @@ class YAMLEditable(VCardWrapper):
         :returns: the parsed datastructure
         :rtype: dict
         """
-        yaml_parser = yaml.YAML(typ='base')
+        yaml_parser = YAML(typ='base')
         # parse user input string
         try:
             contact_data = yaml_parser.load(input)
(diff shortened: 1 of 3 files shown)


skowalak:
This PR references (#276)
The entire `YAMLEditable.to_yaml()` function is rewritten to use `ruamel.yaml`

What I have not done (yet):
* Pretty printing still relies on `helpers.convert_to_yaml()`. I don't know if that level of indentation/alignment is possible with ruamel
* Values will be inserted in the standard template. I don't know if this breaks some workflows involving custom new user templates (#146).
* Helper tests still use `helpers.convert_to_yaml()`. I will add some tests for the new methods when I have some more time.

khard/helpers/__init__.py
```diff
@@ -61,6 +62,101 @@ def get_random_uid() -> str:
...
+      2. list with only one item become this item
+      3. multiline strings use the YAML literal style:
+         https://yaml.org/spec/1.2.2/#literal-style
+    
+    :param value: the value to be sanitized
+    :returns: the sanitized value
+    """
+    # special case for empty values
+    if not value:
+        value = None
```
lucc, line 78: You do not need the local variable `value` (or rather overwrite the argument), you can just `return` directly.

Also see https://pheanex.github.io/pylint/

khard/helpers/__init__.py
```diff
@@ -61,6 +62,101 @@ def get_random_uid() -> str:
...
+    elif isinstance(value, str):
+        if "\n" in value:
+            value = LiteralScalarString(value)
+
+    return value
+
+
+def yaml_dicts(
+        data: Optional[Dict[str, Any]], 
+        defaults: Union[Dict[str, Any], List[str]] = None
```
lucc, line 92: Shouldn't that be a `Optional[Union[...]]` if it can be `None?

khard/helpers/__init__.py
```diff
@@ -61,6 +62,101 @@ def get_random_uid() -> str:
...
+    """
+    format a dict according to template, if empty use specified defaults
+
+    :param data: dict of contact data with keys as types and values as values.
+    :param defaults: default dict to be used if data is empty.
+    :returns: dict of types and values.
+    """
+    if not data:
+        if isinstance(defaults, list):
+            return dict((key, None) for key in defaults)
```
lucc, line 103: You could use `dict.fromkeys` or at least  literal dict comprehension with `{ ... }` like below,  to keep your code style uniform.

khard/helpers/__init__.py
```diff
@@ -61,6 +62,101 @@ def get_random_uid() -> str:
...
+       
+        return defaults
+
+    data_dict = {}
+    for t, nlist in sorted(data.items(), key=lambda k: k[0].lower()):
```
lucc, line 108: Why t and nlist and not key, value?  What do these names stand for, I don't understand them?

khard/helpers/__init__.py
```diff
@@ -61,6 +62,101 @@ def get_random_uid() -> str:
...
+    
+    return data_dict
+
+
+def yaml_addresses(addresses: Optional[Dict[str, Any]],
+                     address_properties: List[str],
+                     defaults: List[str] = []
+                    ) -> Dict[str, Any]:
+    if not addresses:
+        address_fields = dict((key, None) for key in address_properties)
```
lucc, line 119: `dict.fromkeys`

khard/helpers/__init__.py
```diff
@@ -61,6 +62,101 @@ def get_random_uid() -> str:
...
+    return address_dict
+
+
+def yaml_anniversary(anniversary: Union[str, datetime],
+                       version: str) -> Optional[str]:
+    if not anniversary:
+        return None
+
+    if isinstance(anniversary, str):
+        string = anniversary
```
lucc, line 138: You can `return` instead of assigning to `string`.

khard/helpers/__init__.py
```diff
@@ -61,6 +62,101 @@ def get_random_uid() -> str:
...
+                    ) -> Dict[str, Any]:
+    if not addresses:
+        address_fields = dict((key, None) for key in address_properties)
+        return {address_type: address_fields for address_type in defaults}
+
+    address_dict = {}
+    for address_type, address in addresses.items():
+        if isinstance(address, list):
+            address = address[0]
+        address_dict[address_type] = (
```
lucc, line 126: outer parents are not needed.

khard/helpers/__init__.py
```diff
@@ -61,6 +62,101 @@ def get_random_uid() -> str:
...
+        return defaults
+
+    data_dict = {}
+    for t, nlist in sorted(data.items(), key=lambda k: k[0].lower()):
+        data_dict[t] = nlist[0] if len(nlist) == 1 else nlist
+    
+    return data_dict
+
+
+def yaml_addresses(addresses: Optional[Dict[str, Any]],
```
lucc, line 114: Please add a doc comment.

lucc (commented):
I did a quick review of the helper code, just to give you an idea.  I will of course do another one at least when you remove the draft state.

lucc:
I think you need not bother with #146 because that issue is still in the discussion and planing phase. You can also leave the pretty printing alone as we just reuse the yaml conversion functions.  We do not officially claim that the output of the pretty printer is valid yaml.  We can thing about changing the pretty printer in another PR.

Please try to run the tests  and mypy (and maybe even pylint) locally because currently  travis integration is broken. (I need help from @scheibler to fix this.)

If you could also write a test for #276 (the string with an @) that would be great!

skowalak:
I have a little difficulty running the tests. When I run `python setup.py test` hangs at the line
```
test_contact_is_found_if_name_matches (test.test_command_line_interface.AddEmail) ... 
```
I have to interrupt the testing via `CTRL-C` because the process' memory usage grows beyond what I consider normal (10GiB). Is that correct? Could my changes have caused this behaviour, if incorrect?

skowalak:
Hi @lucc ,
after some troubleshooting the abovementioned issue, I tried running the tests on the `develop` branch and had the exact same problem (`tests.test_command_line_interface.AddEmail` hangs in an infinite loop consuming more and more memory). I even tried it on a different computer running a different OS.

So apparently my changes did not break the tests.

- [x] I ran mypy
- [x] I ran pylint
- [x] I wrote a test for #276 
- [ ] I successfully ran the tests

Is there anything more I need to do before this PR can be merged?

khard/carddav_object.py
```diff
@@ -1237,180 +1239,53 @@ def to_yaml(self) -> str:
...
-                    strings.append("Birthday : ")
-            elif line.lower().startswith("categories"):
-                strings += helpers.convert_to_yaml(
-                    "Categories", self.categories, 0, 11, True)
-            elif line.lower().startswith("note"):
-                strings += helpers.convert_to_yaml(
-                    "Note", self.notes, 0, 5, True)
-            elif line.lower().startswith("webpage"):
-                strings += helpers.convert_to_yaml(
-                    "Webpage", self.webpages, 0, 8, True)
```
lucc, line 1411: I like this code reduction :)

lucc (changes requested):
You can skip the broken test like this:
```diff
diff --git a/test/test_command_line_interface.py b/test/test_command_line_interface.py
index cac683c..97b5ecf 100644
--- a/test/test_command_line_interface.py
+++ b/test/test_command_line_interface.py
@@ -480,6 +480,7 @@ class Merge(unittest.TestCase):
         self.assertEqual('contact2.vcf', second)


+@unittest.skip("reason for skipping")
 class AddEmail(unittest.TestCase):

     @TmpConfig(["contact1.vcf", "contact2.vcf"])
```

If I do that (*temporarily*!) the other tests all run fine.


I noticed that the birthday field and phone numbers as well as ZIP codes are now wrapped in quotes. Do you know if there is a way to only quote strings that are problematic (like the @ at the beginning)?

khard/helpers/__init__.py
```diff
@@ -61,6 +62,121 @@ def get_random_uid() -> str:
     return ''.join([random.choice(string.ascii_lowercase + string.digits)
                     for _ in range(36)])
 
+def yaml_clean(value: Optional[Union[str, List, Dict[str, Any], Sequence]]
+              ) -> Optional[Union[Sequence, List, str, Dict[str, Any],
```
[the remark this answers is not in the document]
skowalak: I got a mypy error without the Sequence, meaning one of the properties of `YAMLEditable` that are used in `to_yaml` is of a Sequence type. I removed List though, thank you.

skowalak:
@lucc I implemented the changes you requested.
About the quoting, I am not entirely sure. I believe ZIP codes at least are quoted because they are strings consisting of numbers only and ruamel quotes them so they are not interpreted as `tag:yaml.org,2002:int`, which would for example remove leading zeroes. That does not apply to birthdays (as they include dashes), still they are quoted, when according to the yaml spec they do not have be quoted. I had a look over at [ruamels source code](https://sourceforge.net/p/ruamel-yaml/code/ci/default/tree/emitter.py) but could not find out why.

So, if that is a big problem we could ofc ask Anthon van der Neut from ruamel if he knows if it is possible not to quote this.

lucc (approved):

lucc:
thank you  @skowalak !

We can leave the quoting for later. That is just aesthetics in my opinion.  The important thing is that yaml editing should be more robust now that #276 should be fixed.

PR #309, opened 2022-06-12
Add kind column in contact table

afh:
This is a first, likely incomplete, attempt to add a _kind column in [the] contact table_ as mentioned in the [`todo.txt`](https://github.com/lucc/khard/blob/develop/todo.txt#L4).

I haven't been able to get the tests running with my [`nixpkgs`](https://github.com/NixOS/nixpkgs) setup on macOS due to errors (see below). Being unfamiliar with `mypy` any comment that might be helpful is appreciated.

<details><summary><code>% nix-shell -p nsh python3 python3Packages.unidecode python3Packages.ruamel-yaml python3Packages.vobject python3Packages.atomicwrites python3Packages.configobj python3Packages.setuptools python3Packages.pip pylint mypy --run "mypy khard"</code></summary>

```
khard/carddav_object.py:20: error: Library stubs not installed for "atomicwrites" (or incompatible with Python 3.9)
khard/carddav_object.py:20: note: Hint: "python3 -m pip install types-atomicwrites"
khard/carddav_object.py:20: note: (or run "mypy --install-types" to install all missing stub packages)
khard/carddav_object.py:23: error: Skipping analyzing "vobject": module is installed, but missing library stubs or py.typed marker
khard/address_book.py:10: error: Skipping analyzing "vobject.base": module is installed, but missing library stubs or py.typed marker
khard/address_book.py:10: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
khard/address_book.py:10: error: Skipping analyzing "vobject": module is installed, but missing library stubs or py.typed marker
khard/config.py:11: error: Skipping analyzing "configobj": module is installed, but missing library stubs or py.typed marker
khard/config.py:16: error: Skipping analyzing "validate": module is installed, but missing library stubs or py.typed marker
khard/khard.py:288: error: Argument 1 to "append" of "list" has incompatible type "methodcaller"; expected "attrgetter[Any]"
khard/khard.py:290: error: Argument 1 to "append" of "list" has incompatible type "methodcaller"; expected "attrgetter[Any]"
khard/khard.py:445: error: Incompatible types in assignment (expression has type "TermQuery", variable has type "AnyQuery")
khard/khard.py:448: error: Incompatible types in assignment (expression has type "AndQuery", variable has type "AnyQuery")
Found 10 errors in 4 files (checked 14 source files)
```

</details>


afh:
I probably should have made this a draft pull request…

lucc:
# mypy and tests

The tests are also broken on the develop branch, I am working on it in #307. Some of your errors are already fixed in the github-actions branch but when you use nix you have to build a python env with the python packages you want, just installing them side by side does not work. Check https://nixos.org/manual/nixpkgs/stable/#python 

With `nix-shell --pure -p python3.withPackages' (p: with p;[mypy ruamel_yaml vobject configobj unidecode])' --run "mypy khard"` I still get 
```
khard/carddav_object.py:23: error: Skipping analyzing 'vobject': found module but no type hints or library stubs
khard/carddav_object.py:1275: error: Argument 1 to "yaml_clean" has incompatible type "object"; expected "Union[str, Sequence[Any], Dict[str, Any], None]"
khard/address_book.py:10: error: Skipping analyzing 'vobject.base': found module but no type hints or library stubs
khard/address_book.py:10: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
khard/address_book.py:10: error: Skipping analyzing 'vobject': found module but no type hints or library stubs
khard/config.py:11: error: Skipping analyzing 'configobj': found module but no type hints or library stubs
khard/config.py:16: error: Skipping analyzing 'validate': found module but no type hints or library stubs
Found 6 errors in 3 files (checked 14 source files)
```
but with --ignore-missing-imports it is only 
```
khard/carddav_object.py:1275: error: Argument 1 to "yaml_clean" has incompatible type "object"; expected "Union[str, Sequence[Any], Dict[str, Any], None]"
Found 1 error in 1 file (checked 14 source files)
```

# your PR

Firstly: Thank you :)

1. I think you do not need a config option "preferred_kind" as [the standard](https://datatracker.ietf.org/doc/html/rfc6350#section-6.1.4) says 

> Implementations MUST support the specific string values defined
      above.  If this property is absent, "individual" MUST be assumed
      as the default.

And if a config option would actually have been needed it needs to be added to the config specification file.

2. Even if some upstream tests are broken you can still implement tests for your new code.

afh:
Thanks for your feedback, I'll work it in.
Regarding tests it seemed to me that tests did not run at all due to the errors I saw, yet that could well be my unfamiliarity with `mypy`; I'll look into it.

lucc:
mypy does not run the tests.  It is only a type checker. Use `python setup.py test` to run the tests.  You can also have a look at the  travis config file for the exact commands that would be run in CI.

afh:
Thanks for your helpful explanations, @lucc. I went ahead and fixed the tests in regards to the newly added `kind` column being displayed.

My preference would be to merge this PR once you find it acceptable and work on the `kind` editing and filtering in a separate PR (#310), but I'm happy to do what you think works best for you.

ℹ️ I'm uncertain whether the following is an issue only with my setup, yet the `test_contact_is_found_if_name_matches (test.test_command_line_interface.AddEmail) ...` test hangs, i.e. is unresponsive and does not finish; I let it "run" for a minute and decided then to ^C it…

khard/formatter.py
```diff
@@ -18,12 +18,12 @@ class Formatter:
...
     def __init__(self, display: str, preferred_email: List[str],
-                 preferred_phone: List[str], preferred_kind: str,
-                 show_nicknames: bool, parsable: bool) -> None:
+                 preferred_phone: List[str], show_nicknames: bool,
+                 parsable: bool) -> None:
         self._display = display
         self._preferred_email = preferred_email
         self._preferred_phone = preferred_phone
-        self._preferred_kind = preferred_kind
+        self._default_kind = "individual"
```
lucc, line 26: If this is a constant you can move it out of `__init__`

lucc:
I am unsure if we should display the kind attribute by default.  My reasoning goes like this:
- I assume that most users would have only "individual" vcards in their address books
- "individual" is the default if no kind attribute is present on the vcard. This means we add noise to the output if the attribute is not set.
- if only "individial" vcards are in a listing it adds very little info to display this.

I have two ideas to improve this: 
- a config/command line option to show the kind
- a similar logic like with the address book column: only show it if not all cards are "individual" (the address book column is only shown when the output contains cards from more than one address book but I would also show the kind if all cards in the output have the same kind but that kind is not "individual")

What do you think?

khard/formatter.py
afh: Valid point. Although this has already been reworked in #310 (see [`khard/carddav_object.py:60`](https://github.com/lucc/khard/pull/310/files#diff-a189618efcc4fcc11fb726bee1742df072e671aace70782e53e4b2b8c1074f70R60)), so it may make sense to merge the two PRs after all…

afh:
I think your idea to improve displaying the `kind` column conditionally and via a config/command-line option makes sense. I'll look into it. Thanks for the suggestion :)

khard/carddav_object.py
```diff
@@ -448,6 +449,11 @@ def _prepare_birthday_value(self, date: Date) -> Tuple[Optional[str],
                 fmt = "%F"
         return date.strftime(fmt), False
 
+    @property
+    def kind(self) -> str:
+        kind = self._get_string_field("kind") or self._default_kind
+        return kind if kind != "org" else "organisation"
```
afh, line 455: I'm uncertain whether `org` should be displayed as `organisation`…
`khard` requires to use `organisations` when querying for the organisation. Yet the corresponding value of the `KIND` property is defined as `org` (see [RFC 6350 6.1.4 KIND](https://datatracker.ietf.org/doc/html/rfc6350#section-6.1.4)) and the RFC always spells it as `organization` (note the use of `z` instead of `s`).

What are your thoughts on this @lucc

afh:
@lucc, this should be ready for another look.
I went ahead and added a new config option (`show_kinds` similar to `show_uids`) allowing to configure whether the `kind` column should always be shown (defaults to `no`) as well as displaying the `kind` column conditionally whether the result list consists entirely of or contains contacts with non-`individual` kind.

test/test_command_line_interface.py
```diff
@@ -210,6 +210,27 @@ def test_postaddr_lists_only_contacts_with_post_addresses(self):
                   '                             SomeState, HomeCountry']
         self.assertListEqual(expect, text)
 
+    def test_mixed_kinds(self):
+        with TmpConfig(["org.vcf", "individual.vcf"]):
```
lucc, line 214: Did you commit these files? I don't see them.

lucc (changes requested):
This looks good, just the test files seem to be missing.

test/test_command_line_interface.py
afh: I forgot to add them… 🤦‍♂️

khard/carddav_object.py
afh: Sticking with the British spelling as `khard` already uses "organisation" rather than "organization" seems to a good solution for the time being then.

lucc:
Lets move on to #310 :)

afh:
Thanks for merging, @lucc 😃

PR #310, opened 2022-06-12
Allow querying by kind

afh:
This is a first, likely incomplete, attempt to add an _option to filter contact table (--kind)_ as mentioned in the [`todo.txt`](https://github.com/lucc/khard/blob/develop/todo.txt#L5).

~~ℹ️ This PR depends on changes introduced in #309 (merged). Please see #309 for details on why this PR does not (yet) contain tests.~~

Any comments on whether this is headed into the right direction or not is greatly appreciated.

### DIFF
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -1093,6 +1093,10 @@ class YAMLEditable(VCardWrapper):
         self._set_string_list(self._add_organisation, "Organisation",
                               contact_data)
 
+        # kind
+        self._delete_vcard_object("KIND")
+        self.kind = contact_data["Kind"]
+
         # role
         self._delete_vcard_object("ROLE")
         self._set_string_list(self._add_role, "Role", contact_data)
@@ -1427,6 +1431,11 @@ class CarddavObject(YAMLEditable):
                 self._get_additional_names() + self._get_last_names() + \
                 self._get_name_suffixes()
             strings.append("Full name: {}".format(list_to_string(names, " ")))
+
+        # kind
+        if self.kind:
+            strings.append("Kind: {}".format(self.kind))
+
         # organisation
         if self.organisations:
             strings += helpers.convert_to_yaml(
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -356,9 +356,9 @@ class VCardWrapper:
             anniversary.params['VALUE'] = ['text']
             anniversary.value = value
         elif self.version == "4.0":
-            self.vcard.add('anniversary').value = value
+            self.vcard.add('ANNIVERSARY').value = value
         else:
-            self.vcard.add('x-anniversary').value = value
+            self.vcard.add('X-ANNIVERSARY').value = value
 
     def _get_ablabel(self, item: vobject.base.ContentLine) -> str:
         """Get an ABLABEL for a specified item in the vCard.
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -466,7 +466,8 @@ class VCardWrapper:
 
     @kind.setter
     def kind(self, value: str) -> None:
-        self.vcard.add("KIND").value = value
+        attr_prefix = "" if self.version == "4.0" else "X-"
+        self.vcard.add(f'{attr_prefix}KIND').value = value
 
     @property
     def formatted_name(self) -> str:
@@ -1102,7 +1103,7 @@ class YAMLEditable(VCardWrapper):
         try:
             self.kind = contact_data["Kind"]
         except KeyError:
-            self.kind = self._default_kind
+            pass
 
         # role
         self._delete_vcard_object("ROLE")
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -461,13 +461,16 @@ class VCardWrapper:
 
     @property
     def kind(self) -> str:
-        kind = self.get_first("kind", self._default_kind)
+        kind = self.get_first(self._kind_attribute_name().lower(), self._default_kind)
         return kind if kind != "org" else "organisation"
 
     @kind.setter
     def kind(self, value: str) -> None:
-        attr_prefix = "" if self.version == "4.0" else "X-"
-        self.vcard.add(f'{attr_prefix}KIND').value = value
+        value = value if value != "organisation" else "org"
+        self.vcard.add(self._kind_attribute_name()).value = value
+
+    def _kind_attribute_name(self) -> str:
+        return "{}KIND".format("" if self.version == "4.0" else "X-")
 
     @property
     def formatted_name(self) -> str:
@@ -1099,7 +1102,7 @@ class YAMLEditable(VCardWrapper):
                               contact_data)
 
         # kind
-        self._delete_vcard_object("KIND")
+        self._delete_vcard_object(self._kind_attribute_name())
         try:
             self.kind = contact_data["Kind"]
         except KeyError:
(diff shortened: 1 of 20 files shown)


lucc:
The todo file is very old and we have since introduced a query syntax. I would prefer a new query key word `kind:` over the option `--kind`.  This would also integrate effortlessly with all subcommands and not just the `list` command.

You could split the editing of the kind attribute via the yaml file into a separate commit and move it to #309. Then we can discuss just the query syntax stuff here. Or you can even move the whole code to #309 and call it "support the kind attribute" or so.

afh:
Ah I see, the query syntax vs. options was not clear to me. I'll rework this PR accordingly.
I'm curious: What is the motivation to keep deprecated options around instead of introduction a new major version with breaking changes?

lucc:
The motivation is that I simply did not make a release yet.  And I was working with my private undocumented policy to not remove stuff unless it has been marked and documented as "deprecated" for at least one release.

afh:
Thanks for the context, @lucc, on why the options are still around. I went ahead and reworked this PR so that it supports querying by `kind`.
Being still rather unfamiliar with `khard`'s code I may have missed a few things. I'll dig into the code a bit more and see where some tests would be helpful. Appreciate if you could direct towards areas that may need more work in order to fully support the kind attribute.

khard/carddav_object.py
```diff
@@ -1078,6 +1082,13 @@ def update(self, input: str) -> None:
         self._set_string_list(self._add_organisation, "Organisation",
                               contact_data)
 
+        # kind
+        self._delete_vcard_object("KIND")
+        try:
+            self.kind = contact_data["Kind"]
+        except KeyError:
+            self.kind = self._default_kind
```
lucc, line 1090: The rfc says the field *may* be present so I would not write one if no value is given in the yaml input

khard/data/template.yaml
```diff
@@ -2,6 +2,10 @@
 # from the full name below if not given.
 Formatted name : 
 
+# kind (requires vcard >= 4.0)
```
lucc, line 5: There is (to my knowledge) no version >4 and we do not know if and how such a version would support a kind attribute.

```suggestion
# kind (requires vcard 4.0)
```

khard/data/template.yaml
```diff
@@ -2,6 +2,10 @@
...
+# one of: individual, group, org, location, application, device
```
lucc, line 6: where did you get "application" and "device" from? I can not see them in the rfc
```suggestion
# one of: individual, group, org, location
```

khard/khard.py
```diff
@@ -23,7 +23,7 @@
 from .formatter import Formatter
 from .helpers import interactive
 from .helpers.interactive import confirm
-from .query import AndQuery, AnyQuery, OrQuery, Query, TermQuery
+from .query import AndQuery, AnyQuery, OrQuery, Query, TermQuery, parse
```
lucc, line 26: this is never used

lucc (commented):
I did a first review.

There is one important question we have to decide early I think: How to handle "kind" with vcard v3.  The rfc for v3 does not talk about kind.  

I think we can either not support kind for v3 or we should check what the rfc says about custom attributes and then treat it as a custom field on v3 cards. (see the anniversary field for an example)

lucc:
About the tests: I would hope for some more test that specifically edit the kind field. And depending on if we support kind on v3 we could have test for this case too.

afh:
Thanks for the helpful review, @lucc; much appreciated. I went ahead and addressed your comments.

How should `khard` behave if a non-standard value is given for `kind`, e.g. `person`? Should it default to `individual`? Skip processing the `kind` attribute? Warn? Use the given value?

khard/data/template.yaml
lucc: I would reference the rfc in that comment and also at the top of the file.

lucc:
Sorry for the delay. I had another look:

- I checked the RFCs for KIND:application and KIND:device and I think it is ok to support them.
- Can you mention these RFCs in the carddav_object.py file at the top with the other rfcs
- if the vobject library validates the kind property we can just use that (check the birthday and anniversary tests for an example) otherwise I suggest do not allow any unsupported values to be set. We should emit a warning if someone tries. The biggest question is how to handle the case if we edit a vcard that already had an unsupported value for KIND and it was not changed by the user during editing?
- for some more unittests you can have a look how it is done for other attributes in test/test_vcard_wrapper.py
- for editing the kinnd attribute you could have a look at the tests in test/test_yaml.py (UpdateVcardWithYamlUserInput)

Can you also rebase this PR on the current develop branch so the list of commits is cleaned up.

afh:
Thanks for your feedback @lucc, much appreciated 🙏 

- [x] Glad to hear that you think it is ok to support them 🙂
- [x] I've added a brief comment and link to the respective RFCs
- [ ] Will look into how the vobject library handles things in the coming days
- [ ] Will add more unit tests in the coming days
- [ ] Will add more edit tests in the coming days
- [ ] Happy to rebase, yet this forces me to force push the changes onto this branch. Would you be open to squash this PR when merging?

afh:
I thought about this a bit more and with my current understanding of things I'd like to run the following by you, @lucc:

* What if `VCardWrapper` had its own `validate` method, which calls `validate` on its `vcard` property.
* What if `VCardWrapper` remembered the attributes that were updated and uses that information in order to be smart about validating, e.g. skip validation of the `kind` attribute.

Would that kind of approach work? I haven't looked too much into vobject's validation implementation. Are you more knowledgable about it than me and could point me to interesting parts?

lucc:
Having a separate validation logic on VCardWrapper might be an interesting idea. But I think we would have to discuss this in a separate issue.

For now I suggest we just check the value of the kind directly our self and do not bother to understand or wrap what vobject does in this regard.  Like this we can get this PR done and discuss new ideas later on.

lucc:
@afh are you still working on this?  We (@scheibler and I) are planning to make the 0.18 release this year and I originally wanted to include this as we have merged #309.

I just had a look at this and added a kind column to one of my vcards and it did not show up with `khard ed`.  I conclude that we need some more tests for this PR.

If you are working on this and want to help get it into the 0.18 release please also rebase on latest develop.  Please also drop me a note if you can not work on it this month, then we will go forward with the release without this PR.

afh:
Please go ahead with the 0.18 release excluding this PR. I'm uncertain if or when I'll find the time to continue on this PR. I've forgotten the details, yet I think the issue grew to wider circles, i.e. the vobject library.

lucc:
Hello @afh I thought about picking this up. Do you still want to work on this or should I take it from here (which would be fine by me)?  I already took the liberty of resolving the conflict with the main branch.

Do you still have some local commits related to this? You mentioned that you added notes about the rfc above but I can not find them.

afh:
Hi @lucc, thanks for reaching out, I'd be happy if you could take it from here my time and attention is unfortunately required elsewhere.

It's been too long so I don't remember what I was referring to, yet I found the following snippet; hope this helps 🤞 
```
For further details about the application and device values for the kind
field see:
- application: https://tools.ietf.org/html/rfc6473
- device: https://tools.ietf.org/html/rfc6869
```

lucc:
@afh I have fixed the additional test I had prepared and went through our old discussion. I think it  is now implemented as far as we discussed it.

Please test the lastest version from the main branch and report back if something does not work (if I remember you are using nix so you can use the flake in this repo :)

Thank you for your work on this PR!

PR #334, opened 2024-06-21
Avoid circular imports

cjwatson:
https://bugs.debian.org/1069838 reported a test failure due to circular imports.  While I couldn't reproduce this locally, I could reproduce it in a VM provided by the bug reporter, and it seems straightforward enough to avoid the problem.

### DIFF
--- a/khard/carddav_object.py
+++ b/khard/carddav_object.py
@@ -16,19 +16,21 @@ import re
 import sys
 import time
 from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, \
-    TypeVar, Union, Sequence, overload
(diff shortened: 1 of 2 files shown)


lucc:
I am happy with the conditional typing import but the local import bothers me.  I was under the impression that it is good python coding style to put all imports at the top. This breaks it.

I understand that we need to do a real import of the carddav_object module at some point because we really use it. But the motivation is not obvious if I read the code in isolation and any future developer who does not have your specific VM will not notice any problem if they change it again.

So the main question is: Is there *really no way* to reproduce this problem outside of your VM? Is there some environment that we can specify declarative to reproduce this?

Issue #127, opened 2017-03-09
accentuated characters sorted wrong


legrostdg:
khard should strip accents before sorting characters.

    $ khard
    1 Eleanor
    2 Eugene
    3 Zakari
    4 Éric

should be

    $ khard
    1 Eleanor
    2 Éric
    3 Eugene
    4 Zakari

scheibler:
Identical to #126.

PR #338, opened 2025-03-13
Support non-accentuated sorting

suiso67:
`unidecode` introduced by [this commit](https://github.com/lucc/khard/commit/cfa3eb11dbfb55f691eca0e669450592546fe3d3) while sorting contacts makes non-latin names being sorted in an unexpected order. This PR provides new sorting option `unaccentuated_sort` which skips `unidecode` during sorting.

There could be a better way to address this issue, like having options for each languages, but it will make sorting address book with multiple language way more complex, so in my opinion this option would be enough for this purpose.


lucc:
I think this is the wrong approach.  The unidecode library says on its frontpage that it will only work "correctly" for alphabets that are similar to latin. So this is a hack anyways.  No we would introduce a command line option that users have to use explicitly in order to switch between a hacky sort option and the default sort order of python (which is also wrong for international strings).

The correct solution is to sort all stuff correctly and not add any extra option that the user needs to set.

We should use python's builtin locale support and sort with `cmp=locale.strcoll` or something similar.   (there seem to be some other libraries notably ICU and pyuca but as khard is single threaded the builtin locale library might be ok) 

If you want you can tackle this, if I find the time I might also try to do it over the weekend. (PS: thanks for bringing this up)

lucc:
@suiso67 can you try the sort-order brach and see if that fixes your problem? It should use the locale from your operating system so you could also try things like 
~~~sh
LC_COLLATE=korean khard list
LC_COLLATE=en_US.UTF-8 khard list
~~~

suiso67:
Didn't know there's a proper way to handle this. Thank you for going through details.
I did run some test using my personal contacts and it works as I think it should.

This is a screenshot of some dummy contacts using `sort = formatted_name` and `display = formatted_name` options:

It does sort Korean names properly, though I haven't gone exhaustive.
> `ㄱ`ㄴㄷㄹㅁㅂㅅㅇ`ㅈ`ㅊㅋ`ㅌ`ㅍ`ㅎ`

By the way I'm not sure why `LC_COLLATE=en_US.UTF-8` does weird thing to Korean names, though. They're sorted in the order of `thjk(탁홍조김)`. It shouldn't be a problem, but I can't seem to understand what criteria they use, and hard to find references regarding how `LC_COLLATE` works for certain locale. Anyway it does what it should at least for what `LC_COLLATE` is set to.

lucc:
@suiso67 I do not know how the English locale sorts Korean characters. But it might just be by byte or unicode value (I think that is what python does by default).

@legrostdg are you still using khard? Are you able to test if the changes in the sort-order branch are breaking sorting for you again? (I am asking you because I have now removed the workaround that was introduced for your bug report from #127).

lucc:
I have now merged the sort-order branch.  Thanks @suiso67 for bringing this up and sorry for not using your code.

suiso67:
> I have now merged the sort-order branch. Thanks @suiso67 for bringing this up and sorry for not using your code.

Not at all, I learned a new thing. Thanks for the work.

legrostdg:
@lucc Sorry, I did not take the time to test the sort-order branch. I just tested the master branch, and the sort order is right. Thanks!

Issue #203, opened 2019-01-16
Allow autodiscovery of contact books via globbing


wirew0rm:
It would be nice, if khard would support globbing for the adressbook configuration like khal does: [khal-discover](https://khal.readthedocs.io/en/latest/configure.html#exemplary-discover-usage)

In the config file this should look like this:

    [[adressbooks]]
    path = ~/.contacts/*
    type = discover

lucc:
Nice idea, you can try to implement a new address book class that does this.

antonmosich:
@lucc I tried to tackle this, and decided to use a different approach. You can look at it in my [discover branch](https://github.com/lucc/khard/compare/main...antonmosich:khard:discover). Would you be fine with that approach? Then I could document it and open a pull request.

lucc:
@antonmosich yes, please do that.

PR #341, opened 2025-05-01
Add discover addressbook type

antonmosich:
As mentioned in #203 I took a shot at allowing globbing. The way this works is pretty similar to khal, essentially if you set the type option of an addressbook to `discover`, the path is run through `glob.glob` with `recursive = True`, and a separate addressbook is added for every folder that is found.

The first commit fixes a small typo, if preferred, I can also open a separate PR for that.

### DIFF
--- a/khard/config.py
+++ b/khard/config.py
@@ -142,6 +142,15 @@ class Config:
 
     @classmethod
     def _unfold_discover_books(cls, addressbooks: configobj.Section) -> configobj.Section:
+        """Expand globs in path of addressbooks of type "discover"
+
+        This expands all addressbooks of type "discover" into (potentially)
+        multiple addressbooks of type "vdir". The names are automatically generated
+        based on the directory name.
+
+        :param config: the configuration to be changed
+        :returns: the changed configuration with no "discover" addressbooks
+        """
         for section_name, book in addressbooks.copy().items():
             if book["type"] != "discover":
                 continue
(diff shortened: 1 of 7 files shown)


khard/config.py
```diff
@@ -137,6 +139,48 @@ def _validate(config: configobj.ConfigObj) -> configobj.ConfigObj:
...
+        for section_name, book in config["addressbooks"].copy().items():
+            if book["type"] != "discover":
+                continue
+            hits = iglob(os.path.expanduser(book["path"]), recursive=True)
+            dirs = cls._find_leaf_dirs(hits)
+            for bookpath in dirs:
+                bookname = os.path.basename(bookpath)
+                # Make sure our name is unique
+                while bookname in config["addressbooks"]:
+                    bookname += "-1"
```
lucc, line 162: Can you please do this with a counter.  Currently this results in names like `foo`, `foo-1`, `foo-1-1` but I would prefer `foo`, `foo-1`, `foo-2`, ...

```suggestion
                i = 0
                while bookname in config["addressbooks"]:
                    i += 1
                    bookname += f"-{i}"
```

khard/config.py
```diff
@@ -137,6 +139,48 @@ def _validate(config: configobj.ConfigObj) -> configobj.ConfigObj:
             raise ConfigError
         return config
 
+    @classmethod
+    def _unfold_discover_books(cls, config: configobj.ConfigObj) -> configobj.ConfigObj:
```
lucc, line 143: Can you change this function to receive and return a `configobj.Section` (the addressbook section) only instead of the whole `ConfigObj`.  The calling code is then responsible to replace the section in the outer object:
```python
# something like this
config["addressbooks"] = self._unfold_discover_books(config["addressbooks"])
```

That would be a more functional programming style and the type signature would assure the reader that only this section is modified.

lucc (changes requested):
This looks good in general, I have made some proposals. Please have a look.

lucc:
@antonmosich thank you for your work!

antonmosich:
Thank you for reviewing and merging!

Issue #342, opened 2025-07-07
dependency on atomicwrites


emtiu:
Hi, I've been enjoying khard on FreeBSD for a number of years now, but it has recently been deprecated in the FreeBSD Ports collection because of its dependency on `atomicwrites`, which (as the FreeBSD Ports maintainers asssess) _"has been archived on Jul 16, 2022. Upstream suggests os.replace and os.rename instead"_

Is there a chance to have khard without this dependency? I don't have a good idea of how exactly khard uses it, but if the maintainers feel that it's not too big of a refactoring, I could try to create a PR.

lucc:
Atomicwrites is only used in one place in khard. It should be possible to refactor the code to not use the package as a dependency. A merge request is welcome.

emtiu:
When #343 is committed and there is a new release, khard can return into the FreeBSD Ports collection :)

emtiu:
Thank you for the quick work :) Can you tag a release, or is that not practical at the moment?

lucc:
The release is also requested in #344. I am working on it.

PR #343, opened 2025-07-10
No atomicwrites

mathstuf:
---
Fixes: #342.

### DIFF
--- a/khard/contacts.py
+++ b/khard/contacts.py
@@ -1354,7 +1354,11 @@ def atomic_write(dest: str, overwrite: bool = False) -> Iterator[IO[str]]:
     try:
         yield file
     except Exception:
-        os.unlink(src)
+        try:
+            file.close()
+            os.unlink(src)
+        except Exception:
+            pass
         raise
     else:
         file.flush()


lucc (changes requested):
Looking good so far.  

Can you please add a docstring and type hints like in the rest of the code.

If you have the time you could also migrate the pytest tests from the original atomicwrites repo to stdlib unittest framework. But I recognize that this is a bit more work. (Test would be nice though so if you do not have the time I will look into it)

And I added a commit which I would ask you to squash into the "revert" commit because I want to keep some of the nix code changes.

mathstuf:
I added some docstrings. I don't really know what typehints do with `@contextlib.contextmanager`, so I'll just use CI to figure that out…

mathstuf:
I didn't drag the tests over; I don't really have a modern Python dev set up to iterate on that rapidly.

mathstuf:
Not as a `with` statement at least. Making a context manager to handle exceptions and error handling logic keeps it tidy even with a single call site, IMO.

mathstuf:
To be more precise, open-coding it would push the actual "meat" of the logic in by 2 more indentations and decorate it with a dozen or so lines of "safety handling" noise (at least as far as reviewing the call site logic is concerned).

emtiu:
I see, thanks for the explanation!

PR #349, opened 2026-01-15
FIX: Special handling for ambiguous date formats

### LINKED ISSUE #335: DeprecationWarning on python3.11
`khard ls <name>` leads to:

```
/usr/bin/khard:8: DeprecationWarning: Parsing dates involving a day of month without a year specified is ambiguious
and fails to parse leap day. The default behavior will change in Python 3.15
to either always raise an exception or to use a different default year (TBD).
To avoid trouble, add a specific year to the input & format.
See https://github.com/python/cpython/issues/70647.
  sys.exit(main())

<khard responce>
```

TroyDowling:
This small patch revises the `string_to_date` function in `typing.py` to more gracefully handle parsing dates without a year. The motivation being to address the DeprecationWarning issued by Python in recent versions (and to avoid a possible breakage in the upcoming Python 3.15).

This closes #335 on the bug board.

* Interpolates the placeholder year 1900 for the two ambiguous formats.
* Formatted with `ruff`.

Before
```
................./home/troy/projects/khard/khard/helpers/typing.py:76: DeprecationWarning: Parsing dates i
nvolving a day of month without a year specified is ambiguous
and fails to parse leap day. The default behavior will change in Python 3.15
to either always raise an exception or to use a different default year (TBD).
To avoid trouble, add a specific year to the input & format.
See https://github.com/python/cpython/issues/70647.
  return datetime.strptime(string, fmt)
.................../home/troy/projects/khard/khard/cli.py:208: PendingDeprecationWarning: FileType is depr
ecated. Simply open files after parsing arguments.
  type=argparse.FileType("w"),
..........................................................................................................
....................................................................................ss....................
.....................................................................x....................................
.............................................
----------------------------------------------------------------------
Ran 399 tests in 0.326s

OK (skipped=2, expected failures=1)
```

After
```
..................................../home/troy/projects/khard/khard/cli.py:208: PendingDeprecationWarning:
 FileType is deprecated. Simply open files after parsing arguments.
  type=argparse.FileType("w"),
..........................................................................................................
....................................................................................ss....................
.....................................................................x....................................
.............................................
----------------------------------------------------------------------
Ran 399 tests in 0.327s

OK (skipped=2, expected failures=1)
```

and

```
$ python -W error::DeprecationWarning -m unittest test.test_helpers_typing.StringToDate
..........
----------------------------------------------------------------------
Ran 10 tests in 0.003s

OK

```

### DIFF
--- a/khard/contacts.py
+++ b/khard/contacts.py
@@ -33,7 +33,7 @@ from khard.exceptions import Cancelled
 from . import address_book  # pylint: disable=unused-import # for type checking
 from . import helpers
 from .helpers.typing import (Date, PostAddress, StrList, convert_to_vcard,
-    list_to_string, string_to_date, string_to_list)
+    list_to_string, string_to_date, string_to_list, DEFAULT_YEAR)
 from .query import AnyQuery, Query
 
 
@@ -450,7 +450,7 @@ class VCardWrapper:
                 return date.strip(), True
             return None, False
         tz = date.tzname()
-        if date.year == 1900 and date.month != 0 and date.day != 0 \
+        if date.year == DEFAULT_YEAR and date.month != 0 and date.day != 0 \
                 and date.hour == 0 and date.minute == 0 and date.second == 0 \
                 and self.version == "4.0":
             fmt = '--%m%d'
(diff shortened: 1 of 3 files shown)


aeneby:
Wouldn't this break for Feb 29th, though? Since 1900 was not a leap year.

aeneby:
To be honest, I didn't realise khard already uses 1900 as a placeholder year, and actually so does Python.

I kind of like @lucc's suggestion (https://github.com/lucc/khard/issues/335#issuecomment-2567085640)
around extending the datetime class to accommodate yearless dates. This would be more robust than making assumptions about a specific year being a placeholder value.

PR #355, opened 2026-06-10
Fix bug skipping config's skip_unparsable setting

aadityadhruv:
`skip_unparsable` always has a default value of False, both in the config spec and in the CLI args. If the config sets it to `true`, and if the user does not pass the CLI flag, the merge function will overwrite the default CLI false and NOT skip the unparsable vcard, despite it being set to yes in the config. By adding a logical OR with `config.skip_unparsable`, we can correct the merge behavior

A thing to note here is that `config.skip_unparsable` is always defined since it is created when `self._set_attributes()` is called.

### DIFF
--- a/khard/cli.py
+++ b/khard/cli.py
@@ -395,7 +395,15 @@ def parse_args(argv: list[str]) -> tuple[argparse.Namespace, Config]:
 
     # Save the last option that needs to be carried from the first parser run
     # to the second.
-    skip = args.skip_unparsable
+    # The default value for skip_unparsable in both the parser, and the config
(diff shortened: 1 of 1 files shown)


lucc:
I do not like this solution because it does mix the logic of `parse_args` with `merge_args_into_config`.  I also found that we already hat tests for this but they where broken.  So I fixed and extended the tests and fixed the bug you are also fixing here by setting a proper default value of None on the command line options (for when they are unset).

Thank you for the report any ways and sorry that I did not take your code.

aadityadhruv:
Thanks for the fix!

